PR #43, opened 2024-04-15
[controller] Add LVM thin provisioning setting and switch to distroless

duckhawk:
## Description

1. Switched images to distroless
2. Added LVM thin provisioning setting

We want  to have disable LVM thin provisioning possibility

Added LVM thin provisioning setting
More secure images

## Checklist
- [ ] e2e tests passed.


images/webhooks/src/validators/mcValidator.go
```diff
@@ -0,0 +1,110 @@
...
+			klog.Fatal(err.Error())
+		}
+
+		objs := dhctl.ModuleConfigList{}
+
+		err = cl.List(ctx, &objs)
+
+		for _, obj := range objs.Items {
+			if obj.Name == "sds-node-configurator" {
+				if obj.Spec.Settings == nil {
```
AleksZimin, line 94: Do we need to add a version field?

AleksZimin (approved):

PR #90, opened 2024-08-29
[controller] Add BlockDevice resource selector for LVMVolumeGroups

ViktorKram:
## Description
1. The `LvmVolumeGroup` kind is renamed to `LVMVolumeGroup` kind.
2. Added the `spec.blockDeviceSelector` field to `LVMVolumeGroup` resources.
3. Removed the `spec.blockDeviceNames` field from `LVMVolumeGroup` resources (now you specify the `BlockDevice` resources through the selector).
4. Added the `spec.local` field.

1. The `LvmVolumeGroup` kind is renamed to `LVMVolumeGroup` kind.
This is needed for better code style and code consistency. To achieve this, the controller will provide a migration to replace the old `LvmVolumeGroup` CRD with the new `LVMVolumeGroup` one. Every `LvmVolumeGroup` resource will be migrated as well, changing its kind to `LVMVolumeGroup` with new spec functionality.

2. `BlockDeviceSelector` field in `LVMVolumeGroup` resource's spec.
This feature allows users to specify a selector for `BlockDevice` resources, which will be used to configure the `LVMVolumeGroup`. The selector field replaces the old `spec.blockDeviceNames` field, providing an opportunity to fully configure the `LVMVolumeGroup` resources while bootstrapping a cluster. It is also a more flexible way to operate with `BlockDevice` resources for the `LVMVolumeGroup` ones.

3. Added `spec.local` field.
This field includes only one subfield `nodeName`, which stores the information about the node the resource is placed on.

1. lvmvolumegroups CRD and CRs have LVMVolumeGroup kind
2. Old-kind LvmVolumeGroup resources are migrated to new-kind LVMVolumeGroup ones (they have Ready state, the information in them is correct, and their `spec.blockDeviceNames` are stored as `spec.blockDeviceSelector`.

## Checklist
- [x] e2e tests passed.


.github/workflows/go_modules_check.yaml
```diff
@@ -37,34 +37,34 @@ jobs:
           echo "Processing $go_mod_file"
           
           while IFS= read -r line; do
-            if [[ "$line" == *github.com/deckhouse/sds-* || "$line" == *github.com/deckhouse/csi-* || "$line" == *github.com/deckhouse/virtualization ]]; then
+            if [[ ("$line" == *github.com/deckhouse/sds-* || "$line" == *github.com/deckhouse/csi-* || "$line" == *github.com/deckhouse/virtualization) && "$line" =~ *replace* ]]; then
```
astef, line 40: `=~` -  это regular expression matching и в нём `*` - это часть регулярки, а не wildcard.

```bash
#!/bin/bash

line=qwertyxxreplacexx

if [[ "$line" =~ *replace* ]]; then
    echo MATCHED!
else
    echo NOT MATCHED!
fi

if [[ "$line" =~ replace ]]; then
    echo MATCHED!
else
    echo NOT MATCHED!
fi

if [[ "$line" = *replace* ]]; then
    echo MATCHED!
else
    echo NOT MATCHED!
fi

# Output:
# NOT MATCHED!
# MATCHED!
# MATCHED!
```

astef: Ревью поинт следующий:

- неясно зачем это изменение (добавить в описание)
- почему используется другой стиль матчинга по сравнению с соседними условиями
- в регулярке ошибка

crds/lvmvolumegroupbackup.yaml
```diff
@@ -0,0 +1,286 @@
...
+metadata:
+  name: lvmvolumegroupbackups.storage.deckhouse.io
+  labels:
+    heritage: deckhouse
+    module: storage
+spec:
+  group: storage.deckhouse.io
+  scope: Cluster
+  names:
+    kind: LvmVolumeGroupBackup
```
astef, line 13: А тут точно `LvmVolumeGroupBackup`, а не `LVMVolumeGroupBackup` ?

docs/USAGE.md
```diff
@@ -67,40 +67,58 @@ There are two ways to create an `LVMVolumeGroup` resource:
...
+        nodeName: "node-0"
+      blockDeviceSelector:
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values:
+          - dev-07ad52cef2348996b72db262011f1b5f896bb68f
+          - dev-e90e8915902bd6c371e59f89254c0fd644126da7
+    matchLabels:
+      kubernetes.io/hostname: node-0
```
AleksZimin, line 85:
```suggestion
```

AleksZimin: I think it would be confusing for users. `matchLabels` doesn't make any sense in this case. We should add a separate example with `matchLabels` without the kubernetes.io/metadata.name expression.

docs/USAGE.md
```diff
@@ -67,40 +67,58 @@ There are two ways to create an `LVMVolumeGroup` resource:
...
+        nodeName: "node-0"
+      blockDeviceSelector:
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values:
+          - dev-07ad52cef2348996b72db262011f1b5f896bb68f
+          - dev-e90e8915902bd6c371e59f89254c0fd644126da7
+        matchLabels:
+          kubernetes.io/hostname: node-0
```
AleksZimin, line 108:
```suggestion
```

docs/USAGE_RU.md
```diff
@@ -67,40 +67,58 @@ description: Использование и примеры работы конт
...
+        nodeName: "node-0"
+      blockDeviceSelector:
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values:
+          - dev-07ad52cef2348996b72db262011f1b5f896bb68f
+          - dev-e90e8915902bd6c371e59f89254c0fd644126da7
+    matchLabels:
+      kubernetes.io/hostname: node-0
```
AleksZimin, line 85:
```suggestion
```

docs/USAGE_RU.md
```diff
@@ -67,40 +67,58 @@ description: Использование и примеры работы конт
...
+        nodeName: "node-0"
+      blockDeviceSelector:
+        matchExpressions:
+        - key: kubernetes.io/metadata.name
+          operator: In
+          values:
+          - dev-07ad52cef2348996b72db262011f1b5f896bb68f
+          - dev-e90e8915902bd6c371e59f89254c0fd644126da7
+        matchLabels:
+          kubernetes.io/hostname: node-0
```
AleksZimin, line 108:
```suggestion
```

docs/USAGE_RU.md
```diff
@@ -117,6 +135,9 @@ description: Использование и примеры работы конт
 kubectl delete lvg %lvg-name%
 ```
 
+### Вывод ресурса `BlockDevice` из `LVMVolumeGroup` ресурса
+Для того чтобы вывести `BlockDevice` ресурс из `LVMVolumeGroup` ресурса, необходимо либо изменить поле `spec.blockDeviceSelector` `LVMVolumeGroup` ресурса (добавить другие селекторы), либо изменить соответствующие лейблы у `BlockDevice` ресурса, чтобы они больше не попадали под селекторы `LVMVolumeGroup`. 
+
```
AleksZimin, line 140:
```suggestion
### Вывод ресурса `BlockDevice` из `LVMVolumeGroup` ресурса
Для того чтобы вывести `BlockDevice` ресурс из `LVMVolumeGroup` ресурса, необходимо либо изменить поле `spec.blockDeviceSelector` `LVMVolumeGroup` ресурса (добавить другие селекторы), либо изменить соответствующие лейблы у `BlockDevice` ресурса, чтобы они больше не попадали под селекторы `LVMVolumeGroup`.  После этого Вам необходимо самостоятельно на ноде выполнить команды pvmove, vgreduce и pvremove

```

images/agent/src/pkg/cache/cache.go
```diff
@@ -92,7 +93,7 @@ func (c *Cache) StoreLVs(lvs []internal.LVData, stdErr bytes.Buffer) {
 	}
 
 	for key, lv := range c.lvs {
-		if lv.Exist {
+		if lv.Exist && reflect.ValueOf(lv.Data).IsZero() {
```
astef, line 96: This loop is a cache cleanup procedure, and this condition was keeping non-removed records from being cleaned up. After this change, only zero-valued ones will be kept, which doesn't make sense.

Apart from this, couple of points regarding problems, that were there before:
- Line 100 - condition is excessive and can be removed, since all existent (`Exist: true`) items will be filtered on line 96, and non-existent should be deleted regardless of their presense in `lvsOnNode`
  - this refactor will let us remove `lvsOnNode` variable at all, we'll just use `lvs`
- Line 86 is excessive, because the key is already present as a for loop variable

astef: More complex existing problem, which came to my mind.

Since we're cleaning up removed records, we're losing information about them being removed. This information is helpful to avoid overwriting them with "somehow-late" store operations, see condition on the line 87, where we avoid overwriting `Exist: false` records.

So we'll end up with invalid cache records after some "late" operations. Solution would be to keep the timestamp of the entry removal and do a cleanup only for really old records (with no chance of catching "somehow late update operation")

images/agent/src/pkg/controller/block_device.go
```diff
@@ -162,7 +162,30 @@ func hasBlockDeviceDiff(blockDevice v1alpha1.BlockDevice, candidate internal.Blo
...
+	err = cl.List(ctx, list, &client.ListOptions{LabelSelector: s})
+	metrics.APIMethodsDuration(BlockDeviceCtrlName, "list").Observe(metrics.GetEstimatedTimeInSeconds(start))
+	metrics.APIMethodsExecutionCount(BlockDeviceCtrlName, "list").Inc()
+	if err != nil {
+		metrics.APIMethodsErrors(BlockDeviceCtrlName, "list").Inc()
+		return nil, err
+	}
+
+	result := make(map[string]v1alpha1.BlockDevice, len(list.Items))
+	for _, i := range list.Items {
```
astef, line 181: `i` can easily be confused with `index`, can we use name `item` here?

images/agent/src/pkg/controller/block_device.go
```diff
@@ -162,7 +162,30 @@ func hasBlockDeviceDiff(blockDevice v1alpha1.BlockDevice, candidate internal.Blo
...
+		result[i.Name] = i
+	}
+
+	return result, nil
+}
+
+func GetAPIBlockDevices(ctx context.Context, kc client.Client, metrics monitoring.Metrics) (map[string]v1alpha1.BlockDevice, error) {
```
astef, line 188: This method is mostly the same as `GetAPIBlockDevicesBySelector`, we can refactor it into using `GetAPIBlockDevicesBySelector` with `nil` `selector`.

astef (approved):

AleksZimin (approved):

duckhawk (approved):

PR #94, opened 2024-09-18
[controller] Add a BlockDevice labels watcher controller

ViktorKram:
## Description
Add a BlockDeviceLabelsWatcher controller to watch for the `BlockDevice` resources labels, and if some of the resources start to match or no longer match the `LVMVolumeGroup` `spec.blockDeviceSelector`, trigger such `LVMVolumeGroup` resource to update their configurations.

Without the BlockDeviceLabelsWatcher-controller user's LVMVolumeGroup resources will not be able to be updated just in time and might have unexpected or incorrect states.

If a BlockDevice resource starts to match existing LVMVolumeGroup `spec.blockDeviceSelector`, such `BlockDevice` should be added to the LVMVolumeGroup (for example, corresponding VolumeGroup will be extended with a new block device)
If a BlockDevice no longer matches existing LVMVolumeGroup `spec.blockDeviceSelector`, such LVMVolumeGroup resource should got `ConfigurationApplied` false, as the LVMVolumeGroup resource lost its BlockDevice resource.

## Checklist
- [x] e2e tests passed.


images/sds-health-watcher-controller/src/pkg/controller/block_device_labels_watcher.go
```diff
@@ -0,0 +1,182 @@
...
+			for _, d := range n.Devices {
+				usedBdNames[d.BlockDevice] = struct{}{}
+			}
+		}
+		if lvg.Labels == nil {
+			lvg.Labels = make(map[string]string)
+		}
+
+		shouldTrigger := false
+		switch selector.Matches(labels.Set(blockDevice.Labels)) {
```
astef, line 140: Is there any particular reason for `switch-case` instead of `if` here?

AleksZimin (approved):

images/sds-health-watcher-controller/src/pkg/controller/block_device_labels_watcher.go
```diff
@@ -0,0 +1,182 @@
...
+		case false:
+			log.Debug(fmt.Sprintf("[ReconcileBlockDeviceLabels] BlockDevice %s does not match a blockDeviceSelector of the LVMVolumeGroup %s", blockDevice.Name, lvg.Name))
+			if _, used := usedBdNames[blockDevice.Name]; used {
+				log.Warning(fmt.Sprintf("[ReconcileBlockDeviceLabels] the BlockDevice %s does not match the LVMVolumeGroup %s blockDeviceSelector, but is used by the resource. Add the label %s to provide LVMVolumeGroup resource configuration update", blockDevice.Name, lvg.Name, LVGUpdateTriggerLabel))
+				shouldTrigger = true
+				break
+			}
+
+			log.Debug(fmt.Sprintf("[ReconcileBlockDeviceLabels] the BlockDevice %s does not match the LVMVolumeGroup %s blockDeviceSelector and is not used by the resource", blockDevice.Name, lvg.Name))
+		}
```
astef, line 168: Extracting this block into a function `shouldTrigger(...)` could be beneficial for readability. Also it will remove a few lines, since you can always just `return true` from there

images/sds-health-watcher-controller/src/pkg/controller/block_device_labels_watcher.go
```diff
@@ -0,0 +1,188 @@
...
+	}))
+	if err != nil {
+		log.Error(err, "[RunBlockDeviceLabelsWatcher] unable to controller.Watch")
+		return err
+	}
+
+	return nil
+}
+
+func ReconcileBlockDeviceLabels(ctx context.Context, cl client.Client, log logger.Logger, blockDevice *v1alpha1.BlockDevice) (bool, error) {
```
astef, line 103: This function could be made private, it's only used in this package

images/sds-health-watcher-controller/src/pkg/controller/block_device_labels_watcher.go
```diff
@@ -0,0 +1,188 @@
...
+	log.Debug(fmt.Sprintf("[ReconcileBlockDeviceLabels] BlockDevice %s matches a blockDeviceSelector of the LVMVolumeGroup %s", blockDevice.Name, lvg.Name))
+	if _, used := usedBdNames[blockDevice.Name]; !used {
+		log.Info(fmt.Sprintf("[ReconcileBlockDeviceLabels] the BlockDevice %s matches the LVMVolumeGroup %s blockDeviceSelector, but is not used yet. Add the label %s to provide LVMVolumeGroup resource configuration update", blockDevice.Name, lvg.Name, LVGUpdateTriggerLabel))
+		return true
+	}
+
+	// for the case when BlockDevice stopped match the LVG blockDeviceSelector and then start again
+	for _, c := range lvg.Status.Conditions {
+		if c.Type == internal.TypeVGConfigurationApplied && c.Status == metav1.ConditionFalse {
+			log.Warning(fmt.Sprintf("[ReconcileBlockDeviceLabels] the BlockDevice %s matches the LVMVolumeGroup %s blockDeviceSelector, but the LVMVolumeGroup has condition %s in status False. Add the label %s to provide LVMVolumeGroup resource configuration update", blockDevice.Name, lvg.Name, c.Type, LVGUpdateTriggerLabel))
```
astef, line 170: > Add the label...

First I thought it's a recommendation to the reader of the logs. But then I realized it's a statement of what will be done on line 147.

There're already logs about the label on lines 148, 153, so may be let's remove "Add the label" part of the logs here and in two other places nearby

astef (approved):

PR #114, opened 2024-12-19
Move LLVS to EE

astef:
## Description
Functionality of snapshots should be available only for EE edition of the product. With this change we're:

- separating EE/CE logical branches in Go code
- adding support of conditional test/lint/build to out CI
- letting developer choose the edition with labels 
- adding validating webhook, which will forbit creating of LLVS
- LLVS will be present in all editions 

This change aligns the existing feature design with Deckhouse policies.

EE users will not be affected
CE users will not be able to create LLVS, existing LLVSs won't be reconciled, PVC with LLVS source will lead to controller errors.

## Checklist
- [X] Changes were tested in the Kubernetes cluster manually.


astef:
This merge was a mistake, it was reverted, and the new [PR created](https://github.com/deckhouse/sds-node-configurator/pull/126)

PR #130, opened 2025-02-26
[controller] Fix LVMVolumeGroup spec.blockDeviceSelector update Issue

AleksZimin:
## Description

This PR addresses an issue where the `spec.blockDeviceSelector` field of an existing `LVMVolumeGroup` resource was not updating during reconciliation, even when new physical volumes (PVs) were manually added on the node to the corresponding LVM Volume Group. With this fix, the `spec.blockDeviceSelector` now correctly reflects all associated `BlockDevice` resources, ensuring accurate representation of the volume group's state.

Previously, when new `BlockDevice` resources were manually introduced to an existing LVM Volume Group on a node, the `spec.blockDeviceSelector` field in the corresponding `LVMVolumeGroup` resource remained unchanged. This discrepancy could lead to mismanagement of storage resources and potential inconsistencies in volume group configurations. By ensuring that `spec.blockDeviceSelector` is updated appropriately, we maintain consistency between the actual state of the LVM Volume Group and its representation within Kubernetes.

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/agent/internal/controller/lvg/discoverer.go
+++ b/images/agent/internal/controller/lvg/discoverer.go
@@ -480,6 +480,29 @@ func (d *Discoverer) UpdateLVMVolumeGroupByCandidate(
 		return updErr
 	}
 
+	// Persist the spec.BlockDeviceSelector update (if any) first: Status().Update() below
+	// cannot write to spec, and the regular Update() preserves status on the API side, so
+	// touching spec after we've populated the status fields would wipe them. Update()
+	// refreshes lvg.ResourceVersion in place, so the subsequent Status().Update() picks up
+	// the new revision.
+	updatedSelector, err := updateBlockDeviceSelectorIfNeeded(lvg.Spec.BlockDeviceSelector, candidate.BlockDevicesNames)
+	if err != nil {
+		return fmt.Errorf("updating block device selector for LVMVolumeGroup %q: %w", lvg.Name, err)
+	}
+	if updatedSelector != nil {
+		lvg.Spec.BlockDeviceSelector = updatedSelector
+		d.log.Debug(fmt.Sprintf("[UpdateLVMVolumeGroupByCandidate] update LVMVolumeGroup %q spec.blockDeviceSelector", lvg.Name))
+
+		start := time.Now()
+		err = d.cl.Update(ctx, lvg)
+		d.metrics.APIMethodsDuration(DiscovererName, "update").Observe(d.metrics.GetEstimatedTimeInSeconds(start))
+		d.metrics.APIMethodsExecutionCount(DiscovererName, "update").Inc()
+		if err != nil {
+			d.metrics.APIMethodsErrors(DiscovererName, "update").Inc()
+			return fmt.Errorf(`[UpdateLVMVolumeGroupByCandidate] unable to update LVMVolumeGroup spec, name: %q, err: %w`, lvg.Name, err)
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


images/agent/src/internal/controller/lvg/discoverer.go
```diff
@@ -870,6 +882,51 @@ func configureBlockDeviceSelector(candidate internal.LVMVolumeGroupCandidate) *m
...
 }
 
+func updateBlockDeviceSelectorIfNeeded(existedLabelSelector *metav1.LabelSelector, blockDeviceNames []string) (*metav1.LabelSelector, bool) {
+	if existedLabelSelector == nil {
+		return &metav1.LabelSelector{
+			MatchExpressions: []metav1.LabelSelectorRequirement{
+				{
+					Key:      internal.MetadataNameLabelKey,
+					Operator: metav1.LabelSelectorOpIn,
+					Values:   blockDeviceNames,
```
astef, line 892: Here, and on line 923, it's better to make a copy of the slice (`slices.Clone(blockDeviceNames)`), since in the scope of the function we don't know where did `blockDeviceName` came from and whether it's immutable.

images/agent/src/internal/controller/lvg/discoverer.go
```diff
@@ -870,6 +882,51 @@ func configureBlockDeviceSelector(candidate internal.LVMVolumeGroupCandidate) *m
...
+				},
+			},
+		}, true
+	}
+
+	updated := false
+	found := false
+	for i := range existedLabelSelector.MatchExpressions {
```
astef, line 900: Introduce loop variable and use it instead of repeating `existedLabelSelector.MatchExpressions[i]` each time

images/agent/src/internal/controller/lvg/discoverer.go
```diff
@@ -870,6 +882,51 @@ func configureBlockDeviceSelector(candidate internal.LVMVolumeGroupCandidate) *m
...
+		if existedLabelSelector.MatchExpressions[i].Key == internal.MetadataNameLabelKey {
```
astef, line 901: Inverse if statement to reduce nesting level

images/agent/src/internal/controller/lvg/discoverer.go
```diff
@@ -870,6 +882,51 @@ func configureBlockDeviceSelector(candidate internal.LVMVolumeGroupCandidate) *m
...
+
+			for _, bd := range blockDeviceNames {
+				if _, exist := existingValuesMap[bd]; !exist {
+					existedLabelSelector.MatchExpressions[i].Values = append(existedLabelSelector.MatchExpressions[i].Values, bd)
+					existingValuesMap[bd] = struct{}{}
+					updated = true
+				}
+			}
+		}
+	}
```
astef, line 917: Here we assume that every expression has 'In' operator. What if there's a `NotIn` expression?

Also, there are cases, when device matches by labels different from `kubernetes.io/metadata.name` . In such cases we will add expressions unnecessarily.

I propose to use `k8s.io/apimachinery/pkg/labels` package to check if labels actually match.

astef (changes requested):

images/agent/src/internal/controller/lvg/discoverer.go
asergunov: Done

asergunov:
This is a quick fix. We are going to implement proper one.

PR #139, opened 2025-03-31
[internal] Let us use mock

asergunov:
## Description

- commands interface for mocking
- discoverer interface for mocking
- cspell config, minor spelling fixes
- golint pre-commit config

Make it easier to write usinttests

## Checklist
- [ ] e2e tests passed.


astef (approved):

images/agent/internal/utils/commands.go
```diff
@@ -34,7 +34,47 @@ import (
...
+	VGChangeDelTag(vGName, tag string) (string, error)
+	LVChangeDelTag(lv internal.LVData, tag string) (string, error)
+	UnmarshalDevices(out []byte) ([]internal.Device, error)
+	ReTag(ctx context.Context, log logger.Logger, metrics monitoring.Metrics, ctrlName string) error
+}
+
+type commands struct {
+}
+
+func NewCommands() Commands {
```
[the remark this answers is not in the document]
asergunov: @astef linter doesn't like this change at all

images/agent/internal/mock_utils/mock.go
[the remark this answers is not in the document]
asergunov: I've done better

.cspell/terms.txt
```diff
@@ -0,0 +1,52 @@
...
+Readyz
+stretchr
+tdata
+tmeta
+vgchange
+vgextend
+vgremove
+vgtest
+VGUID
+VGUUID
```
astef, line 52: Such tooling should be either local on developer machine, or centralized across all repositories. It would be hard to synchronize it across 15+ repos.

Also, it should be "attached" to our release process somehow, otherwise it's just a random file, like personal IDE settings. What is helpful to one developer is just a garbage to another.

astef (changes requested):

.cspell/terms.txt
asergunov: Agreed. There are ways to put in outside repo, manage multiple repos and even dependencies.

asergunov: This just a place for now where we put word once. If not there every developer should manage his own list.

.cspell.json
```diff
@@ -0,0 +1,23 @@
...
+            "name": "companies",
+            "description": "Company names mentioned",
+            "path": ".cspell/companies.txt"
+        }
+    ],
+    "ignorePaths": [".cspell/"],
+    "allowCompoundWords": true,
+    "ignoreRegExpList": [
+        "[\"\\t^]--?\\w+[\"\\t$]",
+        "[a-zA-Z0-9]{6}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{4}-[a-zA-Z0-9]{6}",
```
astef, line 21: Trailing comma in JSON is a syntax error. Some implementations may tolerate it, though

.cspell.json
```diff
@@ -0,0 +1,23 @@
...
+    ]
+}
```
astef, line 23: Can we choose `cspell.config.yaml` over json? Because it would be helpful to have comments, describing what those regexes are needed for.

Also, why not including $schema? It should give autocomplete information to IDE

astef (changes requested):
Summarizing all discussions regarding cspell, pre-commit, etc. Since integration of new tooling is in progress, let's keep `hack/` folder as our main convention.

- keep `generate_code.sh` , but use new tools inside
- add something like `run_precommit.sh`, to define how we are running that new thing. Installing instructions would be helpful too.
- any other top-level tasks you think will be repeated, but keep those scripts very simple, please

.cspell.json
asergunov: Agreed. Changing to yaml. I've used it just because it was the only way to have name starting with `.`

asergunov: Done

.cspell.json
asergunov: it's yaml now.

asergunov: I have better organized version [here](https://github.com/deckhouse/sds-node-configurator/blob/linters/.cspell.json). That way we could split dictionaries, give names to regexp and so on. It's not final version for sure

asergunov:
@astef I belive it's simpler to use `go generate` everywhere than have scrips. I've added `/hack/readme.md` explaining how to use these tools.

astef:
> @astef I belive it's simpler to use `go generate` everywhere than have scrips. I've added `/hack/readme.md` explaining how to use these tools.

It's simpler to have scripts then both scripts and something else. Don't forget that the root of the repository is "owned" by infrastructure, which is multi-language and very diverse.

Check out what we are aiming for with `hack/` folder:
 - https://github.com/kubernetes/kubernetes/tree/master/hack
 - https://github.com/deckhouse/virtualization/tree/main/hack
 - https://github.com/kubernetes-csi/csi-driver-nfs/tree/master/hack

It's also known as `scripts/`:
 - https://github.com/werf/werf/tree/main/scripts
 - https://github.com/longhorn/longhorn-manager/tree/master/scripts
 - https://github.com/helm/helm/tree/main/scripts

I have a strong opinion on two things:
- existing script `generate_code.sh`, should remain, but it may do different things
- whatever you suggest to be integrated into "normal" development flow should have a similar simple script, like `run_cspell.sh` or `run_all_precommit.sh`

I'm ok if preconditions like `pip install pre-commit` would be described in README, but simply giving a link to their site is not reliable enough.

astef (changes requested):
See https://github.com/deckhouse/sds-node-configurator/pull/139#issuecomment-2796557098

astef (approved):

AleksZimin (approved):

PR #171, opened 2025-09-11
[internal] add metrics

duckhawk:
## Description

This PR adds comprehensive LVM metrics collection to the sds-node-configurator agent. The agent now exposes Prometheus metrics for monitoring LVM Volume Groups, Thin Pools, and Logical Volumes across cluster nodes.

Currently, there is no visibility into LVM storage utilization at the cluster level. Operators cannot easily monitor:
- How much storage space is available in Volume Groups
- Thin pool utilization and potential over-provisioning issues
- Individual logical volume usage patterns

After applying these changes:

1. **PodMonitor resource** should be created in `d8-monitoring` namespace:
   kubectl get podmonitor -n d8-monitoring sds-node-configurator-agent
   2. **Metrics endpoint** should be available via kube-rbac-proxy on port 4202:
   kubectl port-forward -n d8-sds-node-configurator ds/sds-node-configurator 4202:4202
   curl -k https://localhost:4202/metrics
   3. **Prometheus** should successfully scrape metrics from all agent pods. Metrics should appear in Prometheus with prefix `sds_node_configurator_`:
   - `sds_node_configurator_lvm_volume_group_size_bytes{node="...", volume_group="..."}`
   - `sds_node_configurator_lvm_thin_pool_used_percent{node="...", volume_group="...", thin_pool="..."}`
   - etc.

4. Only VGs and LVs belonging to managed LVMVolumeGroup resources should have metrics exposed (non-managed VGs are filtered out).

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/agent/internal/monitoring/monitoring.go
+++ b/images/agent/internal/monitoring/monitoring.go
@@ -540,9 +540,6 @@ func (m *Metrics) UpdateLVMMetrics(vgs []internal.VGData, lvs []internal.LVData,
 // This includes VG size/free and thin pool actual/allocated sizes from the LVG status.
 // Returns collected parsing errors that should be logged by the caller.
 func (m *Metrics) UpdateLVGStatusMetrics(lvgs map[string]v1alpha1.LVMVolumeGroup) []error {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-
 	var errs []error
 
 	currentLVGVGs := make(map[string]bool)
@@ -591,6 +588,9 @@ func (m *Metrics) UpdateLVGStatusMetrics(lvgs map[string]v1alpha1.LVMVolumeGroup
 			m.LVGThinPoolAllocationLimitBytes(lvg.Name, vgName, tp.Name).Set(allocationLimitBytes)
 		}
 	}
+
+	m.mu.Lock()
+	defer m.mu.Unlock()
 
 	// Remove stale LVG VG metrics
 	for key := range m.previousLVGVGs {


asergunov (approved):

templates/agent/daemonset.yaml
```diff
@@ -162,15 +163,11 @@ spec:
...
-        ports:
-          - name: metrics
-            containerPort: 4202
-            protocol: TCP
         env:
           - name: HEALTH_PROBE_BIND_ADDRESS
             value: 127.0.0.1:4228
           - name: METRICS_PORT
-            value: 127.0.0.1:4202
+            value: 127.0.0.1:9000
```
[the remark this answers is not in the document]
duckhawk: Port isn't changed

images/agent/internal/monitoring/monitoring.go
```diff
@@ -149,3 +273,227 @@ func (m Metrics) APIMethodsErrors(controllerName, method string) prometheus.Coun
...
+			m.LVMLogicalVolumeUsedPercent(lv.VGName, lv.LVName).Set(usedPercent)
+		}
+	}
+
+	// Remove metrics for VGs, thin pools, and LVs that no longer exist
+	// We need to reset all metrics and then set only current ones
+	// This is a limitation of Prometheus - we can't easily remove specific label combinations
+	// So we'll keep the metrics but they'll be stale until next update
+	// In practice, this is acceptable as metrics will be updated on next scan
+}
```
astef: Review from opus 4.5, looks valid:

UpdateLVMMetrics (lines 352-481)
- Stale metrics never cleaned up (line 353-357): The function creates currentVGs, currentThinPools, and currentLVs maps to track current resources, but these maps are never used. The existing comment at line 476-480 admits this is a limitation, but there's no actual cleanup logic. When VGs/LVs/thin pools are deleted, their Prometheus metrics will remain forever with stale values.
- Silent parsing errors for DataPercent (lines 420-430): If lv.DataPercent is non-empty but fails to parse, the error is silently ignored and metrics are set to 0. Users might think a thin pool is empty when there's actually a parsing error.
- Silent parsing errors for MetadataPercent (lines 432-438): Same issue as above for metadata percentage.
- Inconsistent fallback on parse failure (lines 455-464): For logical volumes, if DataPercent exists but fails to parse, the metrics are set to 0 instead of falling back to the thick volume behavior (100%). This is inconsistent with the intent of the code.

UpdateLVGStatusMetrics (lines 483-519)
- No stale metrics cleanup (lines 485-486): Same issue as UpdateLVMMetrics - when LVGs are deleted or thin pools are removed from an LVG, their metrics remain forever.
- Empty volume_group label (lines 490-491): If lvg.Spec.ActualVGNameOnTheNode is empty (which can happen before the VG is created on the node), metrics will be created with an empty volume_group label, producing misleading metrics.
- Silent AllocationLimit parsing errors (lines 507-508): If tp.AllocationLimit is non-empty but malformed, the error is silently ignored and default 150% is used, which may not reflect the actual intended configuration.

astef (changes requested):

images/agent/internal/monitoring/monitoring.go
```diff
@@ -96,56 +205,418 @@ func init() {
...
+					// Keep default value of 150%
+				} else {
+					allocationLimitPercent = parsed
+				}
+			}
+			allocationLimitBytes := actualSize * allocationLimitPercent / 100.0
+			m.LVGThinPoolAllocationLimitBytes(lvg.Name, vgName, tp.Name).Set(allocationLimitBytes)
+		}
+	}
+
```
kneumoin: It seems that we need to take the castle only here?
m.mu.Lock()
defer m.mu.Unlock()

PR #209, opened 2026-03-30
feat: implement netlink parser istead of lsblk

fastrapier:
### Why do we need it, and what problem does it solve?
Block device discovery previously relied on calling `lsblk` as an external process on every scan cycle. This added latency, created a dependency on the `lsblk` binary in the container image, and was susceptible to blocking when devices enter D (uninterruptible sleep) state.
This PR replaces the `lsblk`-based read path with a udev/sysfs-native approach: an initial crawl via `go-udev/crawler` populates an in-memory `DeviceMap`, which is then kept up to date by netlink events (`HandleEvent`). The `Snapshot()` method reads device properties directly from sysfs (`/sys/class/block`) and the udev database (`/run/udev/data`), eliminating the need to spawn `lsblk`. Mount points are parsed from `/proc/self/mountinfo`.
LVM commands (`pvs`, `vgs`, `lvs`) remain unchanged — they require direct metadata reads that cannot be replaced by netlink.
### What is the expected result?
- `lsblk` is no longer called in the block device discovery path
- Device properties (size, rotational, hotplug, serial, model, WWN, FSType, PartUUID, parent device) are read from sysfs/udev DB
- The `DeviceMap` is fully re-populated on startup and after netlink monitor errors (re-crawl), preventing stale data
- Existing `BlockDevice` CR discovery and filtering behavior is preserved
- The `internal.Device` struct contract is unchanged — no downstream changes needed
## Checklist
- [ ] e2e tests passed.


szhem (commented):
## Code Review Summary

Overall this is a well-designed migration from `lsblk` to native sysfs/udev. The new `udev` package is cleanly structured, the test coverage is thorough (including fake sysfs and integration tests), and the `DeviceMap` correctly uses `sync.RWMutex` for concurrent access.

| Severity | Count |
|---|---|
| Critical | 0 |
| High | 3 |
| Medium | 4 |
| Low | 2 |
| Nit | 1 |

### Positive Observations

- Architecturally sound replacement of `lsblk` — eliminates external binary dependency and D-state blocking
- Excellent test coverage: unit tests for every component with edge cases and fake sysfs
- Integration test comparing Snapshot output with real `lsblk` — great validation approach
- `MergeEnvWithUdevDB` correctly implements event env priority over udev DB
- `ResolveDeviceType` accurately reproduces util-linux `get_type()` logic
- Closure capture race fix in `runControllersReconcile` is correct
- Serial/model/WWN fallback chain matches lsblk `get_properties_by_udev()`

### Low / Nit (PR-level)

**Low:** `fmt.Printf` in `integration_test.go` `getLsblkDevices()` should be `t.Logf` — `Printf` output is not associated with the test and not visible without `-v`.

**Low:** Variable `procSelfMountInfo` is named as if it points to `/proc/self/mountinfo` but actually holds `/proc/1/mountinfo`. Consider renaming to `procPID1MountInfo` or `hostMountInfoPath`.

**Nit:** All new files have `Copyright 2025` but the PR is from 2026.

images/agent/internal/cache/cache.go
```diff
@@ -57,12 +57,12 @@ func New() *Cache {
 	}
 }
 
-func (c *Cache) StoreDevices(devices []internal.Device, stdErr bytes.Buffer) {
+func (c *Cache) StoreDevices(devices []internal.Device, errs []string) {
 	c.devices = devices
-	c.deviceErrs = stdErr
+	c.deviceErrs = errs
```
szhem, line 62: **Severity: High** — `StoreDevices`/`GetDevices` do not acquire the Cache mutex

Neither `StoreDevices` nor `GetDevices` holds `c.m`. Controller goroutines (bd, lvg) spawned by `runControllersReconcile` call `GetDevices()` concurrently with the select-loop goroutine calling `StoreDevices()`. This is a data race on `c.devices` / `c.deviceErrs`.

This is a pre-existing issue, but this PR adds a new writer (`scanAndRefreshMounts` from the mount scan timer), widening the race window.

**Recommendation:** Add locking here and in `GetDevices`:

```go
func (c *Cache) StoreDevices(devices []internal.Device, errs []string) {
	c.m.Lock()
	defer c.m.Unlock()
	c.devices = devices
	c.deviceErrs = errs
}

func (c *Cache) GetDevices() ([]internal.Device, []string) {
	c.m.RLock()
	defer c.m.RUnlock()
	dst := make([]internal.Device, len(c.devices))
	copy(dst, c.devices)
	return dst, c.deviceErrs
}
```

images/agent/internal/scanner/scanner.go
```diff
@@ -89,72 +95,107 @@ func (s *scanner) Run(
...
+				s.log.Info("[RunScanner] successfully ran the controllers reconcile funcs")
 			})
 
 		case err := <-errChan:
-			log.Error(err, "[RunScanner] Monitor udev event error")
+			s.log.Error(err, "[RunScanner] Monitor udev event error, restarting monitor and re-crawling devices")
 			quit = conn.Monitor(eventChan, errChan, matcher)
+			s.crawlDevices(matcher)
 			timer.Reset(duration)
 			continue
```
szhem, line 147: **Severity: High** — Potential event loss during error recovery

The monitor is restarted **before** `crawlDevices`. While `crawlDevices` blocks the select-loop, the monitor goroutine sends events to the unbuffered `eventChan` and blocks. Events accumulate in the kernel netlink socket buffer, which can overflow and drop events if the crawl takes significant time.

After `FillFromCrawler` sets the map, queued events are processed — but any events dropped by the kernel during the crawl are lost permanently.

**Recommendation:** Either buffer `eventChan` to absorb events during crawl:

```go
eventChan := make(chan netlink.UEvent, 256)
```

Or swap the order — crawl first, then restart the monitor — to minimize the window where events can be lost:

```go
case err := <-errChan:
	s.log.Error(err, "...")
	s.crawlDevices(matcher)
	quit = conn.Monitor(eventChan, errChan, matcher)
```

images/agent/internal/scanner/scanner.go
```diff
@@ -191,129 +236,164 @@ func runControllersReconcile(
...
-	devices, cmdStr, stdErr, err := s.commands.GetBlockDevices(ctx)
-	if err != nil {
-		log.Error(err, fmt.Sprintf("[ScanDevices] unable to scan the devices, cmd: %s", cmdStr))
-		return nil, stdErr, err
+// crawlDevices performs a full udev crawl and replaces the device map contents.
+// Used at startup and after netlink monitor reconnection to recover from
+// potentially lost events.
+func (s *scanner) crawlDevices(matcher netlink.Matcher) {
+	crawlerQueue := make(chan crawler.Device)
+	crawlerErrors := make(chan error)
```
szhem, line 340: **Severity: High** — Potential goroutine leak due to unbuffered `crawlerErrors`

If the crawler goroutine tries to send an error to `crawlerErrors` right as (or after) `crawlerQueue` is closed, the select-loop exits (`done = true`) and `close(crawlerQuit)` is called. But if the crawler goroutine is blocked on sending to `crawlerErrors` (unbuffered), it will never see the quit signal and leak forever.

**Recommendation:** Buffer the error channel:

```go
crawlerErrors := make(chan error, 16)
```

Alternatively, drain `crawlerErrors` after the loop exits before calling `close(crawlerQuit)`.

images/agent/internal/udev/scanner.go
```diff
@@ -0,0 +1,167 @@
...
+	dm.mu.RLock()
+	envsCopy := make(map[string]map[string]string, len(dm.devices))
+	for k, v := range dm.devices {
+		envsCopy[k] = v
+	}
+	dm.mu.RUnlock()
+
+	var errs []string
+	result := make([]internal.Device, 0, len(envsCopy))
+	for devID, env := range envsCopy {
```
szhem, line 104: **Severity: Medium** — Non-deterministic device ordering in Snapshot

Iterating over `envsCopy` (a `map`) produces random ordering on every call. The resulting `[]internal.Device` will have different order each time `fillTheCache` calls `Snapshot()`. Downstream controllers comparing device lists may see spurious diffs and do unnecessary work.

**Recommendation:** Sort the result before returning:

```go
import "sort"

sort.Slice(result, func(i, j int) bool {
	return result[i].DevID < result[j].DevID
})
return result, errs
```

images/agent/internal/scanner/scanner.go
```diff
@@ -191,129 +236,164 @@ func runControllersReconcile(
...
-	ctx, cancel := context.WithTimeout(ctx, cfg.CmdDeadlineDuration)
-	defer cancel()
-	devices, cmdStr, stdErr, err := s.commands.GetBlockDevices(ctx)
-	if err != nil {
-		log.Error(err, fmt.Sprintf("[ScanDevices] unable to scan the devices, cmd: %s", cmdStr))
-		return nil, stdErr, err
+// crawlDevices performs a full udev crawl and replaces the device map contents.
+// Used at startup and after netlink monitor reconnection to recover from
+// potentially lost events.
+func (s *scanner) crawlDevices(matcher netlink.Matcher) {
```
szhem, line 338: **Severity: Medium** — No timeout/context for `crawlDevices`

`crawlDevices` runs synchronously in the select-loop and blocks all event processing. If the crawler hangs (e.g. due to broken sysfs), the entire scanner is stuck indefinitely with no timeout or cancellation mechanism.

**Recommendation:** Accept a `context.Context` and wrap with a timeout:

```go
func (s *scanner) crawlDevices(ctx context.Context, matcher netlink.Matcher) {
	ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
	defer cancel()
	// ...
}
```

images/agent/internal/scanner/scanner.go
```diff
@@ -89,72 +95,107 @@ func (s *scanner) Run(
...
+				s.log.Error(err, "[RunScanner] unable to run controllers reconciliations")
 			}
 
-			log.Info("[RunScanner] successfully ran the controllers reconcile funcs")
+			s.log.Info("[RunScanner] successfully ran the controllers reconcile funcs")
+
+		case <-mountScanTimer.C:
+			s.log.Debug("[RunScanner] periodic mount scan")
+			s.scanAndRefreshMounts()
+			mountScanTimer.Reset(mountScanInterval)
```
szhem, line 175: **Severity: Medium** — Duplicate mount-scanning logic

Mount points are updated in two places:
1. `fillTheCache()` — on every device event / timer
2. `scanAndRefreshMounts()` — every 10 seconds

Both call `ParseMountInfo()` and apply mount points to devices. The error handling logic (setting `MountPoint = "unknown"`) is duplicated. Having two writers to the same cache field makes correctness reasoning harder.

**Recommendation:** Extract mount-point application into a single helper and call it from both places, or consolidate mount scanning into `fillTheCache` only (reducing the timer interval if more frequent updates are needed).

images/agent/internal/scanner/scanner.go
```diff
@@ -171,15 +212,19 @@ func runControllersReconcile(
...
-				bdRes, err = bdCtrl(ctx)
+		go func(res controller.Result) {
+			for res.RequeueAfter > 0 {
+				log.Warning(fmt.Sprintf("[runControllersReconcile] BlockDevices reconcile needs a retry in %s", res.RequeueAfter.String()))
+				time.Sleep(res.RequeueAfter)
+				var retryErr error
+				res, retryErr = bdCtrl(ctx)
+				if retryErr != nil {
+					log.Error(retryErr, "[runControllersReconcile] BlockDevices retry reconcile failed")
+					return
```
szhem, line 223: **Severity: Medium** — Retry goroutine silently gives up on transient errors

When `bdCtrl(ctx)` returns an error during retry, the goroutine logs it and exits with `return`. If the error is transient (e.g. network timeout), no further retry happens, and BlockDevice CRs remain out of sync until the next netlink event.

**Recommendation:** Add backoff retry instead of immediate exit:

```go
if retryErr != nil {
	log.Error(retryErr, "... retry failed, will retry with backoff")
	res.RequeueAfter *= 2
	if res.RequeueAfter > 5*time.Minute {
		log.Error(retryErr, "... giving up after max backoff")
		return
	}
	continue
}
```

images/agent/internal/udev/integration_test.go
```diff
@@ -0,0 +1,124 @@
...
+	out, err := exec.Command("lsblk", "-J", "-lpfb", "-no",
+		"name,MOUNTPOINT,PARTUUID,HOTPLUG,MODEL,SERIAL,SIZE,FSTYPE,TYPE,WWN,KNAME,PKNAME,ROTA").Output()
+	require.NoError(t, err, "lsblk command failed")
+
+	var devices internal.Devices
+	require.NoError(t, json.Unmarshal(out, &devices), "failed to unmarshal lsblk output")
+	t.Logf("lsblk returned %d devices", len(devices.BlockDevices))
+
+	for i, d := range devices.BlockDevices {
+		fmt.Printf("lsblk[%d]: Name=%s KName=%s Type=%s Size=%s\n", i, d.Name, d.KName, d.Type, d.Size.String())
```
szhem, line 120: **Severity: Low** — Use `t.Logf` instead of `fmt.Printf`

`fmt.Printf` writes to stdout directly, so this output is not associated with the test, not visible without `-v`, and interleaves with other test output in parallel runs.

**Recommendation:**
```go
t.Logf("lsblk[%d]: Name=%s KName=%s Type=%s Size=%s", i, d.Name, d.KName, d.Type, d.Size.String())
```

images/agent/internal/udev/device.go
```diff
@@ -0,0 +1,394 @@
...
+	"strconv"
+	"strings"
+)
+
+// Package-level paths; tests may override.
+var (
+	sysClassBlockPath = "/sys/class/block"
+	sysBlockPath      = "/sys/block"
+	runUdevDataPath   = "/run/udev/data"
+	procSelfMountInfo = "/proc/1/mountinfo"
```
szhem, line 32: **Severity: Low** — Misleading variable name `procSelfMountInfo`

The name implies `/proc/self/mountinfo`, but the value is `/proc/1/mountinfo` (PID 1 namespace, needed for `hostPID: true` containers). This will confuse future readers.

**Recommendation:** Rename to `procPID1MountInfo` or `hostMountInfoPath`:
```go
var hostMountInfoPath = "/proc/1/mountinfo"
```

images/agent/internal/udev/device.go
```diff
@@ -0,0 +1,394 @@
+/*
+Copyright 2025 Flant JSC
```
szhem, line 2: **Severity: Nit** — Copyright year

`Copyright 2025` but the PR is from March 2026. All new files in this package have the same year. Consider updating to 2026.

szhem (commented):
Three more inline comments for the Low/Nit findings.

szhem (commented):
One more finding — magic values.

PR #218, opened 2026-04-07
[agent] Fix race condition and reconciliation loop in LVG/LLV controllers

duckhawk:
## Description

Fixes several related issues in the agent's LVG/LLV reconciliation pipeline that caused infinite loops, repeating errors, and stuck LVMVolumeGroup status.

### 1. Race condition in `UpdateLVGConditionIfNeeded` (`client_lvg.go`)

The LVG reconciler and the LVG discoverer (scanner goroutine) both call `Status().Update()` on the same `LVMVolumeGroup` concurrently. When the discoverer updates status between the reconciler's `Get` and `Status().Update()`, the reconciler fails with `"the object has been modified"`. Since the stale object is reused for all subsequent writes, every condition update in the same cycle fails in cascade.

**Fix**: Wrap `UpdateLVGConditionIfNeeded` with `retry.RetryOnConflict`. On each attempt it re-fetches the latest `LVMVolumeGroup`, applies the condition to the fresh copy, and propagates the updated `Status` and `ResourceVersion` back to the caller.

### 2. Infinite pvresize loop (`lvg/reconciler.go`)

The `shouldLVGWatcherReconcileUpdateEvent` predicate compared `round_UP(PVSize)` vs `round_UP(DevSize)`. Due to LVM metadata overhead and PE alignment, `PVSize` is always less than `DevSize`, so aligned values were always different — the predicate returned `true` on every update event. This created an infinite loop: reconciler runs pvresize (no-op) → sets condition to Applied → UpdateFunc fires → predicate returns true → reconciler runs again.

**Fix**: Replace the state-based check (`alignedPV != alignedDev`) with a delta-based check (`oldLVG.Status.Nodes != newLVG.Status.Nodes`). The predicate now triggers reconciliation only when the discoverer actually updates device size data, not when the reconciler itself changes conditions.

### 3. Zero ExtentSize crashes (`llv/reconciler.go`, `llv_extender/reconciler.go`)

`utils.AlignSizeToExtent` panics when `ExtentSize == 0`. This happens when the discoverer has not yet populated `Status.ExtentSize` (e.g. for LVGs created before the field was introduced).

**Fix**: Add early guards — the LLV reconciler requeues, and the LLV extender skips the event.

### 4. Missing ExtentSize in discoverer diff (`lvg/discoverer.go`)

`hasLVMVolumeGroupDiff` did not compare `ExtentSize`, so LVGs with `Status.ExtentSize = 0` were never updated by the discoverer — it didn't detect a difference.

**Fix**: Add `candidate.ExtentSize.Value() != lvg.Status.ExtentSize.Value()` to the diff check.

### 5. Non-local LVG events processed unnecessarily (`lvg/reconciler.go`, `llv_extender/reconciler.go`)

`ShouldReconcileCreate` and `ShouldReconcileUpdate` did not filter by node. Events for LVGs belonging to other nodes were enqueued, dispatched, and reconciled — only to be skipped inside `Reconcile`.

**Fix**: Add `Spec.Local.NodeName` checks to predicates in the LVG reconciler and LLV extender.

### 6. Copy-paste log prefix bugs (`lvg/reconciler.go`)

`resizePVIfNeeded` and `extendVGIfNeeded` logged errors under the `[UpdateVGTagIfNeeded]` prefix.

**Fix**: Use correct `[ResizePVIfNeeded]` and `[ExtendVGIfNeeded]` prefixes.

The agent on a multi-node cluster exhibited:
- Repeating `"the object has been modified"` errors every 5s, preventing LVG status from being applied
- After fixing the race, an infinite pvresize loop (~1 cycle/sec) flooding logs and burning CPU
- `"extent size must be positive, got 0"` errors for LLVs on LVGs where ExtentSize was never populated
- Unnecessary processing of LVG events for non-local nodes on every discoverer cycle

All of these issues are resolved by this PR.

- Condition updates succeed even when the discoverer concurrently modifies the same resource
- No more repeating `"the object has been modified"` error logs
- No more infinite pvresize loop — reconciliation only triggers when device sizes actually change
- No more `"extent size must be positive, got 0"` errors — LLV reconcilers gracefully wait for ExtentSize to be populated
- The discoverer correctly populates `Status.ExtentSize` for LVGs created before the field was introduced
- Non-local LVG events are dropped at the predicate level, reducing log noise and unnecessary work

## Checklist
- [ ] e2e tests passed.

### DIFF
--- a/images/agent/internal/controller/llv/reconciler.go
+++ b/images/agent/internal/controller/llv/reconciler.go
@@ -225,6 +225,11 @@ func (r *Reconciler) Reconcile(
 		return controller.Result{RequeueAfter: r.cfg.VolumeGroupScanInterval}, nil
 	}
 
+	if lvg.Status.ExtentSize.Value() == 0 {
+		r.log.Warning(fmt.Sprintf("[Reconcile] LVMVolumeGroup %s has zero ExtentSize (discoverer has not populated it yet), retry in %s", lvg.Name, r.cfg.VolumeGroupScanInterval.String()))
+		return controller.Result{RequeueAfter: r.cfg.VolumeGroupScanInterval}, nil
+	}
+
 	r.log.Debug(fmt.Sprintf("[Reconcile] tries to add the finalizer %s to the LVMLogicalVolume %s", internal.SdsNodeConfiguratorFinalizer, llv.Name))
 	added, err := r.addLLVFinalizerIfNotExist(ctx, llv)
 	if err != nil {
--- a/images/agent/internal/controller/llv_extender/reconciler.go
+++ b/images/agent/internal/controller/llv_extender/reconciler.go
@@ -130,11 +130,6 @@ func (r *Reconciler) shouldLLVExtenderReconcileEvent(newLVG *v1alpha1.LVMVolumeG
 		return false
 	}
 
-	if !utils.LVGBelongsToNode(newLVG, r.cfg.NodeName) {
-		r.log.Debug(fmt.Sprintf("[RunLVMLogicalVolumeExtenderWatcherController] the LVMVolumeGroup %s should not be reconciled as it does not belong to the node %s", newLVG.Name, r.cfg.NodeName))
-		return false
-	}
-
 	if newLVG.Status.Phase != internal.PhaseReady {
 		r.log.Debug(fmt.Sprintf("[RunLVMLogicalVolumeExtenderWatcherController] the LVMVolumeGroup %s should not be reconciled as its Status.Phase is not Ready", newLVG.Name))
 		return false
--- a/images/agent/internal/controller/llv_extender/reconciler.go
+++ b/images/agent/internal/controller/llv_extender/reconciler.go
@@ -90,13 +90,13 @@ func (r *Reconciler) MaxConcurrentReconciles() int {
 }
 
 // ShouldReconcileUpdate implements controller.Reconciler.
-func (r *Reconciler) ShouldReconcileUpdate(_ *v1alpha1.LVMVolumeGroup, _ *v1alpha1.LVMVolumeGroup) bool {
-	return true
+func (r *Reconciler) ShouldReconcileUpdate(_ *v1alpha1.LVMVolumeGroup, objectNew *v1alpha1.LVMVolumeGroup) bool {
+	return objectNew.Spec.Local.NodeName == r.cfg.NodeName
 }
 
 // ShouldReconcileCreate implements controller.Reconciler.
-func (r *Reconciler) ShouldReconcileCreate(_ *v1alpha1.LVMVolumeGroup) bool {
-	return true
+func (r *Reconciler) ShouldReconcileCreate(obj *v1alpha1.LVMVolumeGroup) bool {
+	return obj.Spec.Local.NodeName == r.cfg.NodeName
 }
 
 // Reconcile implements controller.Reconciler.
--- a/images/agent/internal/controller/llv_extender/reconciler.go
+++ b/images/agent/internal/controller/llv_extender/reconciler.go
@@ -140,6 +140,11 @@ func (r *Reconciler) shouldLLVExtenderReconcileEvent(newLVG *v1alpha1.LVMVolumeG
 		return false
 	}
 
+	if newLVG.Status.ExtentSize.Value() == 0 {
+		r.log.Debug(fmt.Sprintf("[RunLVMLogicalVolumeExtenderWatcherController] the LVMVolumeGroup %s should not be reconciled as its ExtentSize is not populated yet", newLVG.Name))
+		return false
+	}
+
 	return true
 }
 
--- a/images/agent/internal/controller/lvg/discoverer.go
+++ b/images/agent/internal/controller/lvg/discoverer.go
@@ -841,8 +841,8 @@ func hasLVMVolumeGroupDiff(log logger.Logger, lvg v1alpha1.LVMVolumeGroup, candi
 		candidate.VGSize.Value() != lvg.Status.VGSize.Value() ||
 		candidate.VGFree.Value() != lvg.Status.VGFree.Value() ||
 		candidate.VGUUID != lvg.Status.VGUuid ||
-		hasStatusNodesDiff(log, convertLVMVGNodes(candidate.Nodes), lvg.Status.Nodes) ||
-		candidate.ExtentSize.Value() != lvg.Status.ExtentSize.Value()
+		candidate.ExtentSize.Value() != lvg.Status.ExtentSize.Value() ||
+		hasStatusNodesDiff(log, convertLVMVGNodes(candidate.Nodes), lvg.Status.Nodes)
 }
 
 func hasStatusNodesDiff(log logger.Logger, first, second []v1alpha1.LVMVolumeGroupNode) bool {
--- a/images/agent/internal/controller/lvg/reconciler.go
+++ b/images/agent/internal/controller/lvg/reconciler.go
@@ -638,7 +638,7 @@ func (r *Reconciler) shouldLVGWatcherReconcileUpdateEvent(oldLVG, newLVG *v1alph
 		return true
 	}
 
-	if !reflect.DeepEqual(oldLVG.Status.Nodes, newLVG.Status.Nodes) {
+	if hasStatusNodesDiff(r.log, oldLVG.Status.Nodes, newLVG.Status.Nodes) {
 		r.log.Debug(fmt.Sprintf("[shouldLVGWatcherReconcileUpdateEvent] update event should be reconciled as the LVMVolumeGroup %s status nodes have changed", newLVG.Name))
 		return true
 	}
@@ -693,7 +693,7 @@ func (r *Reconciler) syncThinPoolsAllocationLimit(ctx context.Context, lvg *v1al
 	}
 
 	if updated {
-		fmt.Printf("%+v", lvg.Status.ThinPools)
+		r.log.Trace(fmt.Sprintf("[syncThinPoolsAllocationLimit] LVMVolumeGroup %s ThinPools: %+v", lvg.Name, lvg.Status.ThinPools))
 		r.log.Debug(fmt.Sprintf("[syncThinPoolsAllocationLimit] tries to update the LVMVolumeGroup %s", lvg.Name))
 		err = r.cl.Status().Update(ctx, lvg)
 		if err != nil {
--- a/images/agent/internal/controller/lvg/reconciler.go
+++ b/images/agent/internal/controller/lvg/reconciler.go
@@ -1094,7 +1094,7 @@ func (r *Reconciler) resizePVIfNeeded(ctx context.Context, lvg *v1alpha1.LVMVolu
 				if isApplied(lvg) {
 					err := r.lvgCl.UpdateLVGConditionIfNeeded(ctx, lvg, v1.ConditionFalse, internal.TypeVGConfigurationApplied, internal.ReasonUpdating, "trying to apply the configuration")
 					if err != nil {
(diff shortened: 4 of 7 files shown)


images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -634,16 +638,9 @@ func (r *Reconciler) shouldLVGWatcherReconcileUpdateEvent(oldLVG, newLVG *v1alph
...
-	for _, n := range newLVG.Status.Nodes {
-		for _, d := range n.Devices {
-			alignedPV, _ := utils.AlignSizeToExtent(d.PVSize, extentSize)
-			alignedDev, _ := utils.AlignSizeToExtent(d.DevSize, extentSize)
-			if alignedPV.Value() != alignedDev.Value() {
-				r.log.Debug(fmt.Sprintf("[shouldLVGWatcherReconcileUpdateEvent] update event should be reconciled as the LVMVolumeGroup %s PV size is different to device size", newLVG.Name))
-				return true
-			}
-		}
+	if !reflect.DeepEqual(oldLVG.Status.Nodes, newLVG.Status.Nodes) {
```
szhem: **Severity: High** — `reflect.DeepEqual` on `resource.Quantity`

`reflect.DeepEqual` on structs containing `resource.Quantity` is unreliable. `resource.Quantity` has internal state (`Format`, `d`, `i`, `s`) — two quantities with the same value can fail `DeepEqual` if created via different paths (e.g., `resource.MustParse("4Mi")` vs `resource.NewQuantity(4194304, resource.BinarySI)`).

In the informer cache context where both objects are deserialized from the same API server encoding, the risk of false positives is low but non-zero.

The codebase already has `hasStatusNodesDiff` in `discoverer.go` (line 842) that correctly compares Nodes element-by-element with value semantics.

**Recommendation:** reuse `hasStatusNodesDiff` instead of `reflect.DeepEqual`:

```go
if hasStatusNodesDiff(r.log, oldLVG.Status.Nodes, newLVG.Status.Nodes) {
    r.log.Debug(...)
    return true
}
```

This would require exporting the function or moving it to a shared package, but ensures correct value-based comparison of Quantity fields.

images/agent/internal/repository/client_lvg.go
```diff
@@ -73,51 +74,58 @@ func (lvgCl *LVGClient) UpdateLVGConditionIfNeeded(
...
 		}
-	} else {
-		lvgCl.log.Debug(fmt.Sprintf("[updateLVGConditionIfNeeded] no conditions were found in the LVMVolumeGroup %s. Append the condition %s in the end", lvg.Name, conType))
-		lvg.Status.Conditions = append(lvg.Status.Conditions, newCondition)
-	}
 
-	lvgCl.log.Debug(fmt.Sprintf("[updateLVGConditionIfNeeded] tries to update the condition type %s status %s reason %s message %s of the LVMVolumeGroup %s", conType, status, reason, message, lvg.Name))
-	return lvgCl.cl.Status().Update(ctx, lvg)
+		lvg.Status = fresh.Status
+		lvg.ResourceVersion = fresh.ResourceVersion
```
szhem, line 131: **Severity: High** — Full `Status` overwrite may discard caller's in-memory modifications

After a successful `Status().Update(ctx, fresh)`, the entire `Status` from the fresh copy is written to the caller's `lvg`. If the caller modified other Status fields in memory before calling `UpdateLVGConditionIfNeeded` (but hadn't persisted them), those changes are silently lost.

Analysis of current call sites shows this is currently safe — `UpdateLVGConditionIfNeeded` is called at points where the caller doesn't hold unpersisted status modifications. However, this is a fragile implicit contract.

**Recommendation:** Add a doc comment to the function signature documenting this side-effect:

```go
// UpdateLVGConditionIfNeeded updates a condition on the LVMVolumeGroup with retry on conflict.
// IMPORTANT: on success, lvg.Status and lvg.ResourceVersion are replaced with server-side
// values. Callers must not hold unsaved in-memory Status modifications before calling this method.
```

images/agent/internal/repository/client_lvg.go
```diff
@@ -73,51 +74,58 @@ func (lvgCl *LVGClient) UpdateLVGConditionIfNeeded(
...
-				exist = true
-				lvgCl.log.Debug(fmt.Sprintf("[updateLVGConditionIfNeeded] a condition %s was found in the LVMVolumeGroup %s at the index %d", conType, lvg.Name, i))
+		exist := false
+		if len(fresh.Status.Conditions) > 0 {
+			lvgCl.log.Debug(fmt.Sprintf("[updateLVGConditionIfNeeded] there are some conditions in the LVMVolumeGroup %s. Tries to find a condition %s", fresh.Name, conType))
+			for i, c := range fresh.Status.Conditions {
+				if c.Type == conType {
+					if checkIfEqualConditions(c, newCondition) {
+						lvgCl.log.Debug(fmt.Sprintf("[updateLVGConditionIfNeeded] no need to update condition %s in the LVMVolumeGroup %s as new and old condition states are the same", conType, fresh.Name))
+						return nil
```
szhem, line 109: **Severity: Medium** — Early return does not update `ResourceVersion`

When the condition is already in the desired state, the function returns `nil` without updating `lvg.ResourceVersion` or `lvg.Status`. If the discoverer has meanwhile updated Status (and ResourceVersion), the caller continues with a stale ResourceVersion. The next `Status().Update` on this object will fail with a conflict.

This isn't critical since the retry logic in `UpdateLVGConditionIfNeeded` will catch the conflict on the _next_ call, but it adds one unnecessary retry cycle.

**Recommendation:** Even on early return, propagate the fresh state:

```go
if checkIfEqualConditions(c, newCondition) {
    lvgCl.log.Debug(...)
    lvg.Status = fresh.Status
    lvg.ResourceVersion = fresh.ResourceVersion
    return nil
}
```

images/agent/internal/controller/llv_extender/reconciler.go
```diff
@@ -90,13 +90,13 @@ func (r *Reconciler) MaxConcurrentReconciles() int {
...
+func (r *Reconciler) ShouldReconcileUpdate(_ *v1alpha1.LVMVolumeGroup, objectNew *v1alpha1.LVMVolumeGroup) bool {
+	return objectNew.Spec.Local.NodeName == r.cfg.NodeName
 }
 
 // ShouldReconcileCreate implements controller.Reconciler.
-func (r *Reconciler) ShouldReconcileCreate(_ *v1alpha1.LVMVolumeGroup) bool {
-	return true
+func (r *Reconciler) ShouldReconcileCreate(obj *v1alpha1.LVMVolumeGroup) bool {
+	return obj.Spec.Local.NodeName == r.cfg.NodeName
 }
```
szhem, line 100: **Severity: Medium** — Two concerns with the node-check predicates

**1. Inline duplication of node-check logic**

The `lvg/reconciler.go` package uses `checkIfLVGBelongsToNode(obj, r.cfg.NodeName)` for the same check, but since it's private in the `lvg` package, it can't be called from here. The inline `obj.Spec.Local.NodeName == r.cfg.NodeName` duplication risks divergence if ownership logic evolves (e.g., shared VG support).

**Recommendation:** Export a spec-based utility function next to the existing `utils.LVGBelongsToNode`:

```go
// utils/utils.go
func LVGSpecBelongsToNode(lvg *v1alpha1.LVMVolumeGroup, nodeName string) bool {
    return lvg.Spec.Local.NodeName == nodeName
}
```

**2. Inconsistent filtering: `Spec.Local.NodeName` vs `Status.Nodes`**

These predicates now filter by `Spec.Local.NodeName`, but `shouldLLVExtenderReconcileEvent` (line 136) still filters by `Status.Nodes` via `utils.LVGBelongsToNode`. These are two different membership mechanisms (spec = intent, status = discoverer state). For a freshly created LVG, the predicate passes but `shouldLLVExtenderReconcileEvent` rejects. This is defense-in-depth but creates confusing behavior during debugging.

**Recommendation:** Align the check — either replace `utils.LVGBelongsToNode` in `shouldLLVExtenderReconcileEvent` with the spec-based check, or remove it since the predicate already guarantees locality.

images/agent/internal/controller/llv/reconciler.go
```diff
@@ -225,6 +225,11 @@ func (r *Reconciler) Reconcile(
 		return controller.Result{RequeueAfter: r.cfg.VolumeGroupScanInterval}, nil
 	}
 
+	if lvg.Status.ExtentSize.Value() == 0 {
+		r.log.Warning(fmt.Sprintf("[Reconcile] LVMVolumeGroup %s has zero ExtentSize (discoverer has not populated it yet), retry in %s", lvg.Name, r.cfg.VolumeGroupScanInterval.String()))
+		return controller.Result{RequeueAfter: r.cfg.VolumeGroupScanInterval}, nil
+	}
```
szhem, line 231: **Severity: Low** — Good defensive guard

Correct approach — guarding against zero `ExtentSize` prevents a panic in `AlignSizeToExtent`. The requeue strategy is appropriate since the discoverer will eventually populate the field.

Combined with the `ExtentSize` diff check added to `hasLVMVolumeGroupDiff` in `discoverer.go`, this forms a complete solution: the discoverer detects the diff and populates ExtentSize, after which the LLV reconciler can proceed.

images/agent/internal/controller/lvg/discoverer.go
```diff
@@ -841,6 +841,7 @@ func hasLVMVolumeGroupDiff(log logger.Logger, lvg v1alpha1.LVMVolumeGroup, candi
 		candidate.VGSize.Value() != lvg.Status.VGSize.Value() ||
 		candidate.VGFree.Value() != lvg.Status.VGFree.Value() ||
 		candidate.VGUUID != lvg.Status.VGUuid ||
+		candidate.ExtentSize.Value() != lvg.Status.ExtentSize.Value() ||
```
szhem, line 844: **Severity: Nit** — Correct fix

Good catch — without this check, LVGs with `Status.ExtentSize = 0` (created before the field was introduced) would never be updated by the discoverer, leaving the zero value permanently. This completes the ExtentSize fix chain together with the guards in `llv/reconciler.go` and `llv_extender/reconciler.go`.

szhem (commented):
## Code Review Summary

| Severity | Count |
|---|---|
| Critical | 0 |
| High | 2 |
| Medium | 4 |
| Low | 1 |
| Nit | 2 |

### Positive Observations

- **Excellent PR description**: each of the 6 fixes has a thorough root cause analysis and clear explanation of the solution. One of the best PR descriptions I've seen.
- **Idiomatic `retry.RetryOnConflict`**: using the standard Kubernetes utility for conflict handling is the right approach. Propagating `ResourceVersion` and `Status` back to the caller prevents cascading errors.
- **Latent bug fix**: the old condition search loop had no `break` — with duplicate condition types, `index` would be overwritten. The new code correctly adds `break`.
- **Defense in depth**: filtering by `Spec.Local.NodeName` at the predicate level complements existing checks inside `Reconcile`, reducing workqueue load.
- **Well-structured commits**: 5 separate commits, each with a clear scope — convenient for bisect and cherry-pick.
- **Correct pvresize predicate replacement**: switching from an always-true size-based check to a delta-based `Status.Nodes` diff is the right architectural choice, eliminating the root cause of the infinite loop.

---

**Pre-existing issue (not part of this PR)**: `fmt.Printf("%+v", lvg.Status.ThinPools)` at line 667 of `reconciler.go` in `syncThinPoolsAllocationLimit` — debug output left in production code, writing to stdout without log level. Should be replaced with `r.log.Trace(...)` or removed.

**Medium — Tests**: PR explicitly marks `[ ] The code is covered by unit tests.` — no tests written. For changes fixing a race condition, an infinite loop, and crash guards, the lack of tests creates regression risk. At minimum, tests should cover:
- `UpdateLVGConditionIfNeeded` with a mock client returning `Conflict` on the first attempt and `nil` on the second
- `shouldLVGWatcherReconcileUpdateEvent` returning `false` when only conditions changed (Nodes identical)
- Zero ExtentSize guard: reconciler requeues, extender skips

szhem (approved):

PR #221, opened 2026-04-08
chore: remove lsblk remnants after udev/netlink transition

fastrapier:
## Summary
- Remove unused `LSBLKCmd` constant from `const.go`
- Remove `lsblk.dynamic` binary and `libsmartcols.so.1` from agent image build (`werf.inc.yaml`)
- Remove `UnmarshalDevices` from `Commands` interface, implementation, mock and its dedicated tests (replaced with direct `json.Unmarshal` in affected test files)
- Update stale comments referencing lsblk in `daemonset.yaml` and `regression1_test.go`

## Context
After PR #209 replaced the production `GetBlockDevices()` call (which ran `lsblk -J`) with udev/netlink-based `DeviceMap.Snapshot()`, several lsblk artifacts remained. This PR cleans them up.

### What is kept (intentionally)
- **Test fixtures** (`testdata/lsblk_*.json`) -- valid device data used by regression tests, format-agnostic
- **Integration test** (`TestSnapshotMatchesLsblk`) -- cross-validates udev results against system lsblk
- **E2E test** (`runLsblkViaDirectSSH`) -- independent host-level verification of agent-discovered data
- **CHANGELOG** entries -- historical records
- **`device.go` comment** -- documents that `ReadSysfsHotplug` mirrors lsblk HOTPLUG behavior

## Test plan
- [x] `go vet` passes on all changed packages
- [x] No compilation errors (pre-existing `unsafe.Pointer` warning in unrelated file)
- [x] CI build passes
- [ ] Agent image builds without `lsblk.dynamic`


images/agent/internal/controller/bd/discoverer_test.go
```diff
@@ -538,7 +539,10 @@ func TestBlockDeviceCtrl(t *testing.T) {
 
 	t.Run("validateTestLSBLKOutput", func(t *testing.T) {
 		d := setupDiscoverer()
-		devices, err := utils.NewCommands().UnmarshalDevices(testLsblkOutput)
+		var parsed internal.Devices
+		err := json.Unmarshal(testLsblkOutput, &parsed)
+		assert.NoError(t, err)
+		devices := parsed.BlockDevices
 		if assert.NoError(t, err) {
```
szhem, line 546: **Severity: Low** — Duplicate `assert.NoError` check

Line 542 (`assert.NoError(t, err)`) and line 544 (`if assert.NoError(t, err)`) both check the same `err`. The standalone assertion on line 542 is redundant — the `if` guard on line 544 already asserts and gates the subsequent `assert.Equal`.

**Recommendation:** Remove the standalone `assert.NoError` and keep only the `if` guard:

```go
var parsed internal.Devices
err := json.Unmarshal(testLsblkOutput, &parsed)
devices := parsed.BlockDevices
if assert.NoError(t, err) {
    assert.Equal(t, 31, len(devices))
}
```

Alternatively, use `require.NoError` to fail fast and drop the `if` entirely:

```go
var parsed internal.Devices
require.NoError(t, json.Unmarshal(testLsblkOutput, &parsed))
assert.Equal(t, 31, len(parsed.BlockDevices))
```

szhem (commented):
## Code Review: PR #221 — remove lsblk remnants

Clean removal of `lsblk` artifacts after the udev/netlink transition. All touchpoints covered: constant, interface methods, mock methods, tests, build config, shared libraries, comments.

### Summary

| Severity | Count |
|---|---|
| Critical | 0 |
| High | 0 |
| Medium | 0 |
| Low | 1 |
| Nit | 0 |

### Notes on the broader feature branch (`d-studenikin-netlink-parser`)

The following observations apply to the udev implementation in the base branch (not this PR's diff), but are worth considering before the feature branch merges into `main`:

1. **`crawlDevices` blocks event-loop without timeout** — `crawlDevices` is called synchronously in the `Run` goroutine, both at startup and on netlink error recovery. If `crawler.ExistingDevices` hangs, the entire event processing stops. Consider adding `context.WithTimeout` for the recovery path.

2. **Mount-point enrichment duplicated** — the same mountinfo parsing + device enrichment logic exists in both `fillTheCache` and `scanAndRefreshMounts`. Could be extracted into a shared helper.

3. **`mountScanTimer` purpose unclear** — `fillTheCache` already parses `/proc/1/mountinfo` on every call, so the separate 10s mount timer only helps if no block events arrive but mounts change. If this is intentional, a comment explaining the rationale would help.

4. **`procSelfMountInfo` variable name is misleading** — it points to `/proc/1/mountinfo` (PID 1), not `/proc/self/mountinfo`. Consider renaming to `procMountInfoPath` or `hostMountInfoPath`.

5. **`Snapshot()` result order is non-deterministic** — iterating over `map[string]map[string]string` yields different orderings per call. If downstream consumers depend on stable ordering, consider sorting by `DevID`.

### Positive Observations

- Complete and thorough cleanup — no leftover references to `lsblk` or `libsmartcols`
- Test migration preserves assertion coverage while removing the `UnmarshalDevices` dependency
- `regression1_test.go` properly updated with `json.Unmarshal` + Gomega `Expect`

PR #223, opened 2026-04-08
fix(lvg): align thin-pool and PV resize sizing with LVM

IvanOgurchenok:
## Description

This PR fixes LVMVolumeGroup sizing regressions introduced after [`efa5f61c53801579ea84f696f2a81292e3de47c6`](https://github.com/deckhouse/sds-node-configurator/commit/efa5f61c53801579ea84f696f2a81292e3de47c6) removed the old `resizeDelta` tolerance and switched the controller to strict LVM-derived size comparisons.

The change aligns thin-pool percentage sizing, PV resize detection, and validation tolerance with the actual usable LVM sizes reported by `vgs`/`pvs`. It also extends the regression coverage around 100% thin-pool requests, extent rounding, and PV resize behavior. The E2E cluster configuration uses RedOS for the primary nodes so the scenario is exercised on the target OS image.

LVM reports usable PV/VG size after metadata overhead and extent alignment, while the controller still used raw block device sizes in several paths. That mismatch could reject valid percentage-based thin-pool requests, repeatedly trigger `pvresize` on unchanged devices, or fail validation only because the aligned size was rounded by an extent.

After applying this PR:

- percentage-based thin pools such as `99%` and `100%` are calculated from the real created VG size;
- unchanged devices no longer cause repeated `pvresize` attempts caused by the normal LVM metadata gap;
- validation tolerates expected extent rounding for requested thin pools;
- E2E coverage checks that the current disk resize causes a new successful `pvresize` log entry and then stabilizes.

## Checklist

- [x] E2E regression coverage is updated.
- [ ] Full E2E run passed.
- [x] Documentation updated according to the changes. N/A for user-facing docs.


szhem (commented):
(review withdrawn by author)

e2e/tests/sds_node_configurator_test.go
```diff
@@ -715,106 +716,106 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
 
 	Describe("Sds Node Configurator", Ordered, func() {
 		var (
-			testClusterResources *cluster.TestClusterResources
-			e2eCtx               context.Context
-			k8sClient            client.Client
-		)
+		testClusterResources *cluster.TestClusterResources
```
szhem: **Severity: Critical**

**Tier: Hard must — `gofmt` / Go project convention (https://pkg.go.dev/cmd/gofmt)**

This file does not pass `gofmt`. Most of the `+586/-552` in this PR is a one-tab de-indent inside `Describe("Sds Node Configurator", ...)`; the actual functional change (the new PV Resize block) is buried in that noise and the blame/review for the whole file becomes unreadable. Several places have inconsistent depth — for example the opening `var (` here sits one level deeper than the fields and the closing `)`:

```go
Describe("Sds Node Configurator", Ordered, func() {
    var (
    testClusterResources *cluster.TestClusterResources
    e2eCtx               context.Context
    k8sClient            client.Client
)
```

`gofmt -l` confirms it:

```
$ gofmt -l e2e/tests/sds_node_configurator_test.go
e2e/tests/sds_node_configurator_test.go
```

**Failure mode:** CI (`gofmt` / `golangci-lint`) will block the merge, and future reviews of this file will keep stumbling over the mixed indentation.

**Recommendation:** land a separate formatting commit (`gofmt -w e2e/tests/sds_node_configurator_test.go`) and rebase the functional change on top. The final diff should essentially shrink to a single new `It(...)` for PV Resize (see the comment further down).

e2e/tests/sds_node_configurator_test.go
```diff
@@ -825,536 +826,569 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
...
-				err = k8sClient.Create(e2eCtx, lvg)
+			}
+			Expect(tp).NotTo(BeNil(), "thin-pool %q not found in status", thinPoolName)
+			Expect(tp.AllocationLimit).To(Equal(thinPoolAllocationLimit), "thin-pool allocation limit should match spec")
+			Expect(tp.Ready).To(BeTrue(), "thin-pool should be Ready")
+
+			By("✓ LVMVolumeGroup Ready; thin-pool present and Ready; conditions without errors")
+			printLVMVolumeGroupInfo(&created)
+
+				By("=== PV Resize Test ===")
```
szhem: **Severity: High**

**Tier: Strong recommendation — Ginkgo best practice (https://onsi.github.io/ginkgo/#spec-subjects-it) / operational must for observability**

The new PV Resize test is appended inside the existing `It("Should create LVMVolumeGroup with one disk and thin-pool", ...)`. Two problems:

1. **Mixed scope.** If PV Resize fails, Ginkgo reports the failure under the LVG-creation `It`. From the report triage cannot tell whether creation or resize broke, and `--focus` cannot re-run the resize part alone. The `It` title no longer describes what it actually checks.

2. **Weak coverage of the fix.** The test only asserts that `VGSize` grows. The *point* of Bug 3 in `BUG_REPORT_THINPOOL_SIZE.md` is that there is no *infinite* `pvresize` loop and no `PVResizeFailed` condition — neither is verified. A regression in the arithmetic inside `resizePVIfNeeded` would still leave this assertion green as long as the VG grew once.

**Recommendation:** split the resize into a dedicated spec inside the same `Context("LVMVolumeGroup with one disk and thin-pool", Ordered, ...)`, move shared setup (VD, LVG) into `BeforeAll`/`BeforeEach`, and additionally assert:

- no `Conditions[VGConfigurationApplied].Reason == "PVResizeFailed"` at any point;
- `pvs` on the node (same SSH mechanism as `lsblk` used elsewhere in this file) reports the new PV size;
- the `pvresize` invocation count (via `UtilsCommandsExecutionCount` or agent logs) is stable across two samples 30–60s apart.

```go
It("should grow VG after underlying VirtualDisk resize (pvresize)", func() {
    // patch VD, wait for BD size
    // wait for Status.VGSize
    // assert no PVResizeFailed condition
    // sample pvresize counter twice, expect equal
})
```

images/agent/internal/controller/lvg/reconciler_test.go
```diff
@@ -305,6 +305,78 @@ func TestLVMVolumeGroupWatcherCtrl(t *testing.T) {
 			valid, _ := r.validateLVGForUpdateFunc(lvg, bds)
 			assert.False(t, valid)
 		})
+
+		t.Run("with_100_percent_thin_pools_returns_true", func(t *testing.T) {
```
szhem, line 312: **Severity: High**

**Tier: Strong recommendation — a test must fail when the fix is reverted**

This test is claimed to cover the new `allowedFreeSpace` extent tolerance, but it actually exits through the `hasFullThinPool = true` branch, which existed before this PR. Walk-through:

- `first` BD (1G) is already a PV, `second` BD (2G) is new → `additionBlockDeviceSpace = 2G`
- `newTotalVGSize = vg.VGSize (1G) + 2G = 3G`
- thin-pool `100%` → `tpRequestedSize = 3G`
- `AlignSizeToExtent(3G, 4Mi) ≈ 3_003_121_664 ≥ newTotalVGSize = 3G`
- → `hasFullThinPool = true`, the entire `if !hasFullThinPool { allowedFreeSpace := ... }` block is skipped.

If the `+ int64(len(lvg.Spec.ThinPools))*extentSizeVal` line in `validateLVGForUpdateFunc` is removed in a future refactor, this test stays green.

**Recommendation:** add a targeted case that actually enters the `!hasFullThinPool` branch and relies on the tolerance:

- one or two non-full pools (e.g. `50%` each) with sizes chosen so that after `AlignSizeToExtent` they sum to `totalFreeSpace + 1..N * extentSize`;
- assert `valid == true` **and** add a regression comment: *"without the per-pool extent tolerance this assertion must fail"*.

Without such a case the tolerance logic has no enforced contract.

images/agent/internal/controller/lvg/reconciler_test.go
```diff
@@ -452,6 +524,48 @@ func TestLVMVolumeGroupWatcherCtrl(t *testing.T) {
 			valid, _ := r.validateLVGForCreateFunc(lvg, bds)
 			assert.False(t, valid)
 		})
+
+		t.Run("with_100_percent_thin_pools_returns_true", func(t *testing.T) {
```
szhem, line 726: **Severity: High**

**Tier: Strong recommendation — make the test's sensitivity to the fix explicit**

This case *does* exercise `totalThinPoolSize > allowedSpace`, but only by accident: `resource.MustParse("1G")` is decimal (1_000_000_000 B), and `AlignSizeToExtent(2_000_000_000, 4Mi) = 2_000_683_008 > totalVGSize = 2_000_000_000`. Without the `+ N * extentSize` tolerance the test would fail; with it, it passes.

That sensitivity is invisible to the next reader. If anyone ever rewrites the inputs as `1Gi` / `2Gi` (multiples of 4Mi), or renames the helper, the test silently stops validating the fix.

**Recommendation:** keep the decimal sizes but document the intent and add a second case that stresses `len(lvg.Spec.ThinPools) > 1`:

```go
// sizes are intentionally non-multiples of the 4Mi extent:
// AlignSizeToExtent(2G, 4Mi) = 2_000_683_008 > totalVGSize = 2_000_000_000,
// so validation must rely on the per-pool extent tolerance introduced in this PR.
```

An additional case with two thin pools at `50%` each makes the `len(lvg.Spec.ThinPools) * extentSize` term load-bearing and traps regressions that shrink it back to a single extent.

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -1087,7 +1102,19 @@ func (r *Reconciler) resizePVIfNeeded(ctx context.Context, lvg *v1alpha1.LVMVolu
...
+					peStart = pv.PEStart.Value()
+					break
+				}
+			}
+			if peStart == 0 {
+				r.log.Warning(fmt.Sprintf("[ResizePVIfNeeded] unable to find PV %s in the cache or pe_start is 0, skip resizing for now", d.Path))
+				continue
+			}
+
+			if d.DevSize.Value()-(d.PVSize.Value()+peStart) >= extentSize.Value() {
```
szhem, line 1136: **Severity: High**

**Tier: Strong recommendation — arithmetic without unit coverage regresses**

The tightest arithmetic change in this PR — `d.DevSize.Value()-(d.PVSize.Value()+peStart) >= extentSize.Value()` — has no unit coverage. It is only exercised by the new e2e case, which in turn only asserts `VGSize` growth (see the comment in `e2e/tests/sds_node_configurator_test.go` around the new PV Resize block).

**Recommendation:** add three table-driven cases with a mocked `commands.ResizePV`:

1. `DevSize - PVSize - peStart == extentSize - 1` → `ResizePV` must **not** be called.
2. `DevSize - PVSize - peStart == extentSize` → `ResizePV` must be called exactly once (boundary of `>=`).
3. `peStart == 0` (or PV missing from `pvs`) → `ResizePV` must **not** be called, and the skip must be logged (see the comment on line 1112).

Without these, the next refactor of the byte math goes straight to e2e, which is slow and flaky for this class of bugs.

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -792,8 +798,12 @@ func (r *Reconciler) validateLVGForCreateFunc(
...
-		if totalThinPoolSize > totalVGSize.Value() {
-			r.log.Trace(fmt.Sprintf("[validateLVGForCreateFunc] total thin pool size: %s, total vg size: %s", resource.NewQuantity(totalThinPoolSize, resource.BinarySI).String(), totalVGSize.String()))
+		ext := extentSizeForThinPoolAlign(lvg, nil)
+		extentSizeVal := ext.Value()
+		allowedSpace := totalVGSize.Value() + int64(len(lvg.Spec.ThinPools))*extentSizeVal
+
+		if totalThinPoolSize > allowedSpace {
+			r.log.Trace(fmt.Sprintf("[validateLVGForCreateFunc] total thin pool size: %s, allowed space: %d", resource.NewQuantity(totalThinPoolSize, resource.BinarySI).String(), allowedSpace))
 			r.log.Warning(fmt.Sprintf("[validateLVGForCreateFunc] requested thin pool size is more than VG total size for the LVMVolumeGroup %s", lvg.Name))
 			reason.WriteString(fmt.Sprintf("Required space for thin-pools %d is more than VG size %d.", totalThinPoolSize, totalVGSize.Value()))
```
szhem: **Severity: High**

**Tier: Operational must — the user-visible message must match the predicate**

The condition on line 805 compares with `allowedSpace = totalVGSize + len(ThinPools)*extentSize`, but the `reason` reported back to the LVG condition still prints `totalVGSize.Value()`:

```go
reason.WriteString(fmt.Sprintf("Required space for thin-pools %d is more than VG size %d.", totalThinPoolSize, totalVGSize.Value()))
```

**Failure mode:** a user hitting the boundary gets `ValidationFailed` with text like `Required space for thin-pools 2_100_000_000 is more than VG size 2_000_000_000`, while the trace log shows the real comparison was against `allowedSpace = 2_004_194_304`. Support and the customer will argue whether the overrun was 100 MB or 96 MB, and there is no way to tell from the condition alone.

For symmetry, the sibling branch in `validateLVGForUpdateFunc` (around line 950) already says *"more than allowed free space in VG"*, which is correct.

**Recommendation:** print `allowedSpace` and make the tolerance explicit:

```go
reason.WriteString(fmt.Sprintf(
    "Required space for thin-pools %d is more than allowed space %d (VG size %d + extent tolerance %d).",
    totalThinPoolSize, allowedSpace, totalVGSize.Value(), int64(len(lvg.Spec.ThinPools))*extentSizeVal,
))
```

e2e/tests/cluster_config.yml
```diff
@@ -3,7 +3,7 @@ clusterDefinition:
   masters: # Master nodes configuration
     - hostname: "master-1"
       hostType: "vm"
-      osType: "Ubuntu 22.04 6.2.0-39-generic" # See internal/config/images.go
+      osType: "RedOS 8.0 6.6.26-1.red80.x86_64" # See internal/config/images.go
```
szhem: **Severity: High**

**Tier: Operational must — PR description / changelog discipline**

Switching the default e2e OS from Ubuntu 22.04 to RedOS 8.0 is a cluster-wide change that affects every e2e test in this module, not just the new PV Resize case. The PR description does not mention the switch; the reason is only visible in the commit message `89f107c "u 22 not working as well with docker, use redos"`.

**Failure mode:**

- all future PRs are now exercised against RedOS 8.0 by default; any Ubuntu-specific regression is silently lost from CI coverage;
- six months from now nobody will remember which Docker problem on Ubuntu 22.04 caused the switch, or whether it is safe to revert;
- `CHANGELOG` gains no trace of it.

**Recommendation (pick one):**

1. Move this line to a separate PR that documents the exact Docker failure on Ubuntu 22.04, links a reproducer, and (ideally) tracks the return to Ubuntu in an issue.
2. Parametrize `osType` via an environment variable (e.g. `E2E_OS_TYPE`) with Ubuntu as the default; CI jobs that need RedOS set it explicitly.
3. At minimum, extend the PR description and commit message with the concrete error output, and add an entry to `CHANGELOG`.

szhem: **Severity: Low**

The inline comment `# See internal/config/images.go` points to a path that does not exist in this repository — `deckhouse/sds-node-configurator` has no `internal/config/images.go`. It presumably refers to the `storage-e2e` framework. A reader grepping the repo will not find the file.

**Recommendation:** spell out the source repo, e.g. `# See storage-e2e/internal/config/images.go`, or move the list of allowed values into the module's e2e documentation and drop the pointer.

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -1087,7 +1102,19 @@ func (r *Reconciler) resizePVIfNeeded(ctx context.Context, lvg *v1alpha1.LVMVolu
...
 		for _, d := range n.Devices {
-			if d.DevSize.Value()-d.PVSize.Value() > extentSize.Value() {
+			var peStart int64
+			for _, pv := range pvs {
+				if pv.PVName == d.Path {
+					peStart = pv.PEStart.Value()
+					break
+				}
+			}
+			if peStart == 0 {
```
szhem, line 1131: **Severity: Medium**

**Tier: Operational must — SRE workflow / observability**

Two independent events are collapsed into one warning:

```go
if peStart == 0 {
    r.log.Warning(fmt.Sprintf("[ResizePVIfNeeded] unable to find PV %s in the cache or pe_start is 0, skip resizing for now", d.Path))
    continue
}
```

- "PV is not in the cache yet" is normal on the first reconcile after `pvcreate`; it will self-heal in 1–2 cycles but still emits a Warning every loop — log noise, and operators cannot distinguish it from the real problem;
- "PV is in the cache but `pe_start == 0`" should never happen on a healthy LVM2 PV (LVM always reserves ≥ 1 MiB for metadata). If it does, the `pvs` parser or the `-o pe_start` contract broke upstream and the cluster operator must know — but the current log does not tell them so.

**Recommendation:**

```go
var (
    foundPV bool
    peStart int64
)
for _, pv := range pvs {
    if pv.PVName == d.Path {
        foundPV = true
        peStart = pv.PEStart.Value()
        break
    }
}
switch {
case !foundPV:
    r.log.Debug(fmt.Sprintf("[ResizePVIfNeeded] PV %s not in cache yet, skip", d.Path))
    continue
case peStart == 0:
    // Should never happen on a healthy LVM PV; log loud and emit a metric/event.
    r.log.Warning(fmt.Sprintf("[ResizePVIfNeeded] PV %s has pe_start=0 reported by pvs, skipping resize", d.Path))
    continue
}
```

The second branch is also worth wiring to a metric or a Kubernetes Event on the LVG so monitoring picks it up.

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -932,7 +942,12 @@ func (r *Reconciler) validateLVGForUpdateFunc(
 			allocatedSize := getVGAllocatedSize(*vg)
 			totalFreeSpace := newTotalVGSize.Value() - allocatedSize.Value()
 			r.log.Trace(fmt.Sprintf("[validateLVGForUpdateFunc] new LVMVolumeGroup %s thin-pools requested %d size, additional BlockDevices space %d, total: %d", lvg.Name, addingThinPoolSize, additionBlockDeviceSpace, totalFreeSpace))
-			if addingThinPoolSize != 0 && addingThinPoolSize > totalFreeSpace {
+
+			ext := extentSizeForThinPoolAlign(lvg, vg)
+			extentSizeVal := ext.Value()
+			allowedFreeSpace := totalFreeSpace + int64(len(lvg.Spec.ThinPools))*extentSizeVal
```
szhem: **Severity: Medium**

**Tier: Trade-off**

The tolerance is `int64(len(lvg.Spec.ThinPools)) * extentSizeVal`, but `len(lvg.Spec.ThinPools)` includes both already-existing pools and those being added. Alignment rounding-up only happens for new or extended pools — the existing ones have already been accounted for in `allocatedSize` / `totalFreeSpace`.

The formula is therefore slightly wider than strictly necessary. It does not open a real hole (one extent per existing pool is noise in practice), but a future refactor can "tighten" it back to the add-only count and the update tolerance silently disappears.

**Recommendation:** count only the pools that actually contribute to `addingThinPoolSize` in the loop above:

```go
addingPoolsCount := 0
for _, specTp := range lvg.Spec.ThinPools {
    // ...
    if !created || thinPoolSizeDiff > 0 {
        addingPoolsCount++
        addingThinPoolSize += ...
    }
}
...
allowedFreeSpace := totalFreeSpace + int64(addingPoolsCount)*extentSizeVal
```

If the current shape is kept intentionally, drop a one-liner comment explaining why `len(ThinPools)` is preferred over `addingPoolsCount` so the choice survives rebase.

szhem (commented):
## Summary

Reviewing the fixes for Bugs 1–3 from `BUG_REPORT_THINPOOL_SIZE.md` (percentage thin-pool off the real `VGSize`, exact `pe_start` in `resizePVIfNeeded`, per-pool extent tolerance in `validateLVGFor*Func`). The logic of all three fixes is correct and well explained in the bug report and the commit messages. The comments below group the findings that should ideally be addressed before merge.

### Severity overview

| Severity | Count |
|---|---|
| Critical | 1 |
| High | 5 |
| Medium | 6 |
| Low | 2 |
| Nit (PR-wide, below) | 3 |

### PR-wide nits (not attached inline)

- **PR checklist empty.** The "The code is covered by unit tests / e2e tests passed / Documentation updated / Changes were tested manually" boxes are still unchecked. Tick those that apply and mark docs N/A if nothing changed — it's the easiest signal reviewers have that the author walked through the list.
- **Explicit link to `efa5f61`.** The PR description references `resizeDelta` and commit `efa5f61`. GitHub resolves the hash, but a full URL (`https://github.com/deckhouse/sds-node-configurator/commit/efa5f61`) is friendlier in changelogs and IDEs.

### Positive observations

- Each of the three fixes lives in its own focused commit (`9449b2f`, `1cfb2d3`, `e9261c1`) with a clear message.
- Using `pv.PEStart` from `pvs` instead of the heuristic `DevSize - PVSize > extent` is the right side of the LVM contract — `pe_start` is exactly the offset where physical extents begin, and `DevSize - (PVSize + peStart) >= extent` correctly signals unallocated space beyond the current PV.
- `vgSize` in `reconcileLVGCreateFunc` is hoisted out of the per-pool loop; the `vgAfterCreate != nil` branch and the fallback Warning keep the code path visible in logs (see the inline comment on that fallback for a follow-up).
- `PVData.PEStart resource.Quantity` matches the rest of the type and round-trips through `pvs --units B --nosuffix` correctly.
- Table-driven sub-tests `with_100_percent_thin_pools_returns_true` keep the existing test style (see the inline comments on how to make them actually load-bearing).

The rest is style / trade-off and can be discussed inline.

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -792,8 +798,12 @@ func (r *Reconciler) validateLVGForCreateFunc(
 		}
 		r.log.Trace(fmt.Sprintf("[validateLVGForCreateFunc] LVMVolumeGroup %s thin-pools requested space: %d", lvg.Name, totalThinPoolSize))
 
-		if totalThinPoolSize > totalVGSize.Value() {
-			r.log.Trace(fmt.Sprintf("[validateLVGForCreateFunc] total thin pool size: %s, total vg size: %s", resource.NewQuantity(totalThinPoolSize, resource.BinarySI).String(), totalVGSize.String()))
+		ext := extentSizeForThinPoolAlign(lvg, nil)
```
szhem: **Severity: Medium**

**Tier: Strong recommendation — clarity of intent**

`extentSizeForThinPoolAlign(lvg, nil)` here never tries to read the extent size from the cached VG. On a re-create (VG was partially applied and is already in `r.sdsCache`) the value in `lvg.Status.ExtentSize` may not yet be populated either, so the call always falls through to `lvmDefaultPhysicalExtent = 4Mi`. For a fresh LVG that default is fine, but the limitation is not captured anywhere in the code — a reader cannot tell whether this is intentional or an oversight.

**Recommendation:** if it *is* intentional, add a one-line comment:

```go
// VG is not yet created at validate-time; Status.ExtentSize may also be empty → fall back to 4Mi.
extent := extentSizeForThinPoolAlign(lvg, nil)
```

If a cached VG is already available (e.g. recovery after a partial apply), consider:

```go
vg := r.sdsCache.FindVG(lvg.Spec.ActualVGNameOnTheNode) // may be nil
extent := extentSizeForThinPoolAlign(lvg, vg)
```

so the real extent is used when LVM has already reported one.

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -792,8 +798,12 @@ func (r *Reconciler) validateLVGForCreateFunc(
...
+		extentSizeVal := ext.Value()
```
szhem: **Severity: Medium**

**Tier: Strong recommendation — DRY / readability**

`extentSizeForThinPoolAlign(lvg, nil)` is called both on every iteration of the loop above and once more here to build `allowedSpace`, and `extentSizeVal` is used exactly once in the next statement. The same duplication exists in `validateLVGForUpdateFunc` (loop around line 910 + the sibling on line 946).

**Recommendation:** hoist once per call:

```go
extent := extentSizeForThinPoolAlign(lvg, nil)

for _, tp := range lvg.Spec.ThinPools {
    ...
    alignedTpSize, alignErr := utils.AlignSizeToExtent(tpRequestedSize, extent)
    ...
}

allowedSpace := totalVGSize.Value() + int64(len(lvg.Spec.ThinPools))*extent.Value()
```

Folds in the sibling nit about the redundant `extentSizeVal` local.

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -541,8 +541,14 @@ func (r *Reconciler) reconcileLVGCreateFunc(
 		}
 		extentForThinPools := extentSizeForThinPoolAlign(lvg, vgAfterCreate)
 
+		var vgSize resource.Quantity
+		if vgAfterCreate != nil {
+			vgSize = vgAfterCreate.VGSize
+		} else {
+			vgSize = countVGSizeByBlockDevices(blockDevices)
```
szhem: **Severity: Medium**

**Tier: Trade-off**

Happy path (`vgAfterCreate != nil`) now correctly drives the percentage calculation off the real VG, which is exactly the fix for Bug 1. However this `else` branch falls back to `countVGSizeByBlockDevices(blockDevices)` — i.e. to the *same raw sum of `BlockDevice.Status.Size`* that caused the regression — whenever `r.commands.GetVG` fails right after `vgcreate` (transient `nsenter`/lock error, timeout, etc.).

**Failure mode:** `vgcreate` succeeds, `GetVG` fails, the controller picks a thin-pool size off the raw sum; on the next reconcile `validateLVGForUpdateFunc` compares that pool against the actual (smaller) `vg.VGSize` and flips the LVG back to `ValidationFailed`. The Warning is logged, but `kubectl get lvg -o yaml` gives the user nothing to act on.

Additionally, the fact that this is a fallback is not visible in the code itself — the explanation lives only in `BUG_REPORT_THINPOOL_SIZE.md`.

**Recommendation (pick one):**

1. Treat `GetVG` failure as a transient error and return `(true, getErr)` so the reconcile loop comes back and picks up the fresh `VGData` from the cache.
2. Flip `VGConfigurationApplied=False` with `Reason=TransientLVMError` instead of creating a thin-pool on a known-wrong base.
3. At minimum, add a comment right before the `else` explaining that this fallback can re-introduce the regression and why it is still kept.

images/agent/internal/utils/commands.go
```diff
@@ -195,7 +195,7 @@ func (commands) GetLV(vgName, lvName string) (lvData internal.LVData, command st
 
 func (commands) GetAllPVs(ctx context.Context) (data []internal.PVData, command string, stdErr bytes.Buffer, err error) {
 	var outs bytes.Buffer
-	args := []string{"pvs", "-o", "+pv_used,pv_uuid,vg_tags,vg_uuid", "--units", "B", "--nosuffix", "--reportformat", "json"}
+	args := []string{"pvs", "-o", "+pv_used,pv_uuid,vg_tags,vg_uuid,pe_start", "--units", "B", "--nosuffix", "--reportformat", "json"}
```
szhem, line 198: **Severity: Medium**

**Tier: Strong recommendation — pin the external LVM contract**

`pvs --units B --nosuffix -o ...,pe_start` returns `"pe_start":"1048576"` as a string, and `resource.Quantity.UnmarshalJSON` parses it correctly as bytes. But the `pvs -o` column set is an external contract, and a future LVM upgrade (renamed column, reordered output, different suffix handling) would break `PVData.PEStart` only at runtime.

`images/agent/internal/utils/commands_test.go` already has a table-driven test for `GetAllPVs` (around line 156). Extending it to cover `pe_start` is cheap and closes the gap.

**Recommendation:**

```go
{"pv_name":"/dev/vdb", ..., "pv_size":"10G", "pe_start":"1048576"}
```

```go
assert.Equal(t, int64(1048576), got[0].PEStart.Value())
```

Optionally a second row with a non-default `pe_start` (e.g. `2097152`) to make sure the value is not hard-coded anywhere.

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -1087,7 +1102,19 @@ func (r *Reconciler) resizePVIfNeeded(ctx context.Context, lvg *v1alpha1.LVMVolu
 	errs := strings.Builder{}
 	for _, n := range lvg.Status.Nodes {
 		for _, d := range n.Devices {
-			if d.DevSize.Value()-d.PVSize.Value() > extentSize.Value() {
+			var peStart int64
+			for _, pv := range pvs {
```
szhem: **Severity: Low**

The lookup is `O(nodes · devices · pvs)`. In practice these cardinalities are small, so there is no runtime concern, but the loop is easy to flatten and the map can be reused for the `peStart == 0` / `!foundPV` split suggested on line 1112:

```go
peStartByPath := make(map[string]int64, len(pvs))
for _, pv := range pvs {
    peStartByPath[pv.PVName] = pv.PEStart.Value()
}

for _, n := range lvg.Status.Nodes {
    for _, d := range n.Devices {
        peStart, ok := peStartByPath[d.Path]
        if !ok { ... }
        if peStart == 0 { ... }
        ...
    }
}
```

Purely a readability win.

e2e/tests/sds_node_configurator_test.go
IvanOgurchenok: in main was changed. rewrited

e2e/tests/sds_node_configurator_test.go
IvanOgurchenok: Split the PV resize scenario into a dedicated `It(...)` inside the same `Context(..., Ordered, ...)` and kept it on top of the current `main` test layout. The resize spec now checks BlockDevice growth, `pvs` growth on the node, LVG `VGFree` plus PV growth after `pvresize`, absence of `PVResizeFailed`, and stable `pvresize` log count after convergence.

images/agent/internal/controller/lvg/reconciler_test.go
IvanOgurchenok: Added a targeted update-validation case that enters the `!hasFullThinPool` branch and depends on the tolerance logic instead of passing through the existing `100%` shortcut.

images/agent/internal/controller/lvg/reconciler_test.go
IvanOgurchenok: Kept the decimal-size case and documented why `1G/2G` are intentional instead of `1Gi/2Gi`. Also added a second case with two `50%` thin-pools so the per-pool tolerance multiplier is explicitly load-bearing.

images/agent/internal/controller/lvg/reconciler.go
IvanOgurchenok: Added table-driven unit coverage for `resizePVIfNeeded` using the existing mocked `utils.Commands` infrastructure: `< extent`, `== extent`, missing PV, and `pe_start == 0`. E2E still remains the end-to-end proof for the whole resize flow.

images/agent/internal/controller/lvg/reconciler.go
IvanOgurchenok: Updated the create-validation message to report the real predicate: allowed space, raw VG size, and the extent-tolerance component.

images/agent/internal/controller/lvg/reconciler.go
IvanOgurchenok: Split the skip cases in `resizePVIfNeeded`: missing PV in cache is now a quiet retryable debug path, while `pe_start == 0` stays a warning because it should not happen on a healthy LVM PV.

For the follow-up observability question, I intentionally kept this as a warning log only. That matches the current style of this controller better than introducing a new condition/event for an internal diagnostic anomaly; if we want stricter handling later, that likely deserves a small style/rule discussion first.

e2e/tests/cluster_config.yml
IvanOgurchenok: we have a bug with u22 on e2e cluster, so no impact here

IvanOgurchenok: comment from e22 repo, i suppose

images/agent/internal/controller/lvg/reconciler.go
IvanOgurchenok: Tightened update tolerance from `len(lvg.Spec.ThinPools)` to `addingPoolsCount`, so the extra extent allowance matches only pools that are actually being added or expanded.

images/agent/internal/controller/lvg/reconciler.go
IvanOgurchenok: Create-time alignment now uses cached VG extent size when a VG is already available in cache; `4Mi` remains only as the fallback for fresh / not-yet-discovered cases.

images/agent/internal/controller/lvg/reconciler.go
IvanOgurchenok: The functional part of this feedback was addressed together with the cached-VG change on create: the create-validation path now resolves `extent` once per call and reuses it for both alignment and allowed-space calculation. I did not do an extra mechanical cleanup of the whole update path in the first pass, because this was a readability nit rather than part of the bugfix itself.

images/agent/internal/controller/lvg/reconciler.go
IvanOgurchenok: Removed the fallback to raw `BlockDevice` sum after `GetVG` failure. If `GetVG` fails right after create, reconcile now returns a retryable error instead of continuing with a knowingly-wrong size base.

images/agent/internal/utils/commands.go
IvanOgurchenok: Extended `commands_test.go` to cover `pe_start` parsing from `pvs --reportformat json`, including multiple values so the contract is guarded by a regression test.

images/agent/internal/controller/lvg/reconciler.go
IvanOgurchenok: Flattened the PV lookup in `resizePVIfNeeded` to a map keyed by path, which also made the missing-PV vs zero-`pe_start` split more readable.

duckhawk (dismissed):

images/agent/internal/controller/lvg/reconciler.go
```diff
@@ -792,10 +796,14 @@ func (r *Reconciler) validateLVGForCreateFunc(
...
-			r.log.Trace(fmt.Sprintf("[validateLVGForCreateFunc] total thin pool size: %s, total vg size: %s", resource.NewQuantity(totalThinPoolSize, resource.BinarySI).String(), totalVGSize.String()))
+		extentSizeVal := extent.Value()
+		extentTolerance := int64(len(lvg.Spec.ThinPools)) * extentSizeVal
+		allowedSpace := totalVGSize.Value() + extentTolerance
+
+		if totalThinPoolSize > allowedSpace {
+			r.log.Trace(fmt.Sprintf("[validateLVGForCreateFunc] total thin pool size: %s, allowed space: %d", resource.NewQuantity(totalThinPoolSize, resource.BinarySI).String(), allowedSpace))
 			r.log.Warning(fmt.Sprintf("[validateLVGForCreateFunc] requested thin pool size is more than VG total size for the LVMVolumeGroup %s", lvg.Name))
-			reason.WriteString(fmt.Sprintf("Required space for thin-pools %d is more than VG size %d.", totalThinPoolSize, totalVGSize.Value()))
+			reason.WriteString(fmt.Sprintf("Required space for thin-pools %d is more than allowed space %d (VG size %d + extent tolerance %d). ", totalThinPoolSize, allowedSpace, totalVGSize.Value(), extentTolerance))
```
szhem, line 806: **Architectural concern — extent *tolerance* reintroduced into validation.**

These create/update validators gate thin-pool sizing with an extent-sized *tolerance* layered on top of the deterministic `AlignSizeToExtent`:

- create: `allowedSpace = totalVGSize + len(lvg.Spec.ThinPools)*extent` (here)
- update, adding pools: `allowedFreeSpace = totalFreeSpace + addingPoolsCount*extent` (L964–L967)
- update, existing pool: `allowedActualSize = alignedTpSize + extent` (L940)

The extent-alignment refactoring (#201) deliberately replaced the fixed `resizeDelta` tolerance with deterministic up-alignment, precisely to *stop* relying on a slack value. The design rationale explicitly rejected the "just increase the tolerance" alternative, because a tolerance masks real discrepancies and scales with the number of objects.

In the create path `allowedSpace` grows with pool count, so N pools can each overshoot by up to one extent and still pass validation — only to fail later at `lvcreate`. That is the same scaling-slack behavior the refactoring set out to remove.

Question: was a deterministic alternative considered instead of a tolerance — e.g. comparing the sum of *aligned* pool sizes against the (aligned) VG size, or capping each pool at the remaining VG (`min(aligned, remaining)`), mirroring the create path's `alignedTpSize >= vgSize → CreateThinPoolFullVGSpace`? That keeps the request→validate→create round-trip deterministic and avoids reintroducing the slack #201 eliminated. If the tolerance is intentional, a short comment explaining why one extent per pool is the correct, bounded allowance would help future readers.

szhem:
**Scope question: are relative sizes of `LVMLogicalVolume`s in scope here, or only thin-pool (LVG) sizing + PV resize?**

This PR changes the LVG reconciler (thin-pool create/validate), adds `PVData.PEStart`, and reworks PV resize. The `LVMLogicalVolume` paths — `images/agent/internal/controller/llv/reconciler.go` (`reconcileLLVCreateFunc` / `reconcileLLVUpdateFunc`) and `llv_extender/reconciler.go` (`ReconcileLVMLogicalVolumeExtension`) — are not touched. Those were already aligned to extents in #201, so this is **not** a regression.

I'm asking to make the scope explicit: if "relative LV sizes" is also meant to cover percent-sized LLVs (e.g. `lvm-thick` / `lvm-thin` PVCs with `Spec.Size` in `%`), that path is outside this PR's diff and would warrant its own coverage — in particular on a non-default physical extent size (e.g. 16 MiB), where percentage rounding behaves differently.

If the intent is strictly thin-pool sizing + PV resize, this is fine as-is — just flagging so reviewers don't assume the LLV side is covered by this change.

PR #234, opened 2026-04-15
feat(agent): add mountinfo parser for mount-point detection

fastrapier:
## Summary

Adds a parser for Linux mountinfo format as a future replacement for the lsblk `MOUNTPOINT` column when determining whether a block device is mounted.

Currently the agent calls `lsblk -J ... MOUNTPOINT ...` on every scan cycle. This parser reads the same data directly from the kernel (`/proc/1/mountinfo`) without spawning an external process, targeting integration with the netlink event pipeline.

### Why `/proc/1/mountinfo`

The agent's DaemonSet is deployed with [`hostPID: true`](https://github.com/deckhouse/sds-node-configurator/blob/0ef81bae5ae737533dbe79a97832d0600306a213/templates/agent/daemonset.yaml#L76), so PID 1 inside the pod is the host's init process. `/proc/pid/mountinfo` shows mounts in that process's mount namespace ([proc_pid_mountinfo(5)](https://man7.org/linux/man-pages/man5/proc_pid_mountinfo.5.html)), meaning `/proc/1/mountinfo` exposes the host's mount table — the one we need to determine if a block device is mounted on the node.

This is the same PID 1 reference already used by all LVM commands via [`nsenter -t 1 -m`](https://github.com/deckhouse/sds-node-configurator/blob/0ef81bae5ae737533dbe79a97832d0600306a213/images/agent/internal/utils/commands.go#L757), but reading the file directly avoids spawning a separate process.

## What's included

- `ParseMountInfo(path)` — parses any mountinfo-formatted file into `map[major:minor]mountPoint`
- Octal escape decoding (`\040`->space, `\011`->tab, `\012`->newline, `\134`->backslash) for path parity with lsblk
- Last-wins semantics for duplicate devices (bind mounts), matching lsblk MOUNTPOINT behavior ("usually the last mounted instance")
- `major:minor` format validation
- Key is `major:minor` — the canonical kernel device identifier, directly available from netlink `MAJOR`/`MINOR` env vars, no ambiguity with symlinks/device-mapper paths
- Comprehensive tests: multiple entries, duplicate dev ID (last wins), empty file, missing file, short/malformed lines, kernel doc example, all octal escapes, trailing backslash

## Follow-up

The parser will be integrated into the netlink event processing pipeline to populate `Device.MountPoint` without calling lsblk.

### DIFF
--- a/images/agent/internal/utils/mountinfo.go
+++ b/images/agent/internal/utils/mountinfo.go
@@ -23,6 +23,13 @@ import (
 	"strings"
 )
 
+// ProcHostMountInfo is the mountinfo of PID 1 (the host's init process).
+// The agent runs with hostPID: true (see templates/agent/daemonset.yaml), so PID 1
+// in the pod's PID namespace is the host's init. /proc/1/mountinfo therefore
+// exposes the host's mount namespace (see proc_pid_mountinfo(5)).
+// This is equivalent to `nsenter -t 1 -m -- cat /proc/self/mountinfo` (the same
+// PID 1 reference used by all LVM commands, see nsentrerExpendedArgs) but avoids
+// spawning a separate process.
 const ProcHostMountInfo = "/proc/1/mountinfo"
 
 // ParseMountInfo parses a mountinfo file (see proc(5)) into a map of "major:minor" → mount-point path.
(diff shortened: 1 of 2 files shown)


images/agent/internal/utils/mountinfo.go
```diff
@@ -0,0 +1,93 @@
...
+package utils
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+)
+
+const ProcHostMountInfo = "/proc/1/mountinfo"
```
kneumoin, line 33: I might be missing something, but hardcoding /proc/1/mountinfo could be unreliable depending on the container/namespace setup.

kneumoin (dismissed):
Looks good

kneumoin (approved):

PR #235, opened 2026-04-15
e2e: add blockdevice regression scenarios

viktor-karpochev:
## Summary
- add an e2e scenario that verifies a `BlockDevice` is removed after its backing disk disappears
- add an e2e scenario that verifies `LVMVolumeGroup` degrades after replacing the backing block device with a smaller one
- add e2e scenarios for rejecting or reverting manual `BlockDevice` creation and status mutations, and document them in `e2e/README.md`

## Test plan
- [x] `cd e2e && go test ./tests -run TestDoesNotExist -count=0`
- [ ] Run the full e2e smoke suite in CI

Made with [Cursor](https://cursor.com)

### DIFF
--- a/e2e/tests/sds_node_configurator_test.go
+++ b/e2e/tests/sds_node_configurator_test.go
@@ -828,7 +828,7 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
 				Expect(nodeList.Items).NotTo(BeEmpty(), "cluster must have at least one node")
 				realNodeName := nodeList.Items[0].Name
 
-				fakeBDName := e2eFakeBDPrefix + fmt.Sprintf("%d", rand.Intn(100000))
+				fakeBDName := e2eFakeBDPrefix + strconv.Itoa(rand.Intn(100000))
 
 				By(fmt.Sprintf("Step 2: Creating fake BlockDevice %s with nodeName=%s", fakeBDName, realNodeName))
 				fakeBD := &v1alpha1.BlockDevice{
--- a/e2e/tests/sds_node_configurator_test.go
+++ b/e2e/tests/sds_node_configurator_test.go
@@ -46,8 +46,8 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/discovery"
+	"k8s.io/client-go/dynamic"
 	k8sclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
@@ -262,25 +262,13 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
 					forceDeleteAllNonConsumableBlockDevices(e2eCtx, k8sClient, 2*time.Minute)
 				}
 
-				if testClusterResources.BaseKubeconfig == nil || testClusterResources.VMResources == nil {
-					if testClusterResources.BaseKubeconfig == nil {
-						Skip("VirtualDisk creation requires base cluster kubeconfig (Deckhouse virtualization). " +
-							"Set SSH_JUMP_HOST to the base cluster or use TEST_CLUSTER_CREATE_MODE=alwaysCreateNew.")
-					}
-					baseKubeconfig = testClusterResources.BaseKubeconfig
-					clusterVMs = nil
-				} else {
-					clusterVMs = make([]string, 0, len(testClusterResources.VMResources.VMNames))
-					for _, name := range testClusterResources.VMResources.VMNames {
-						if name != testClusterResources.VMResources.SetupVMName {
-							clusterVMs = append(clusterVMs, name)
-						}
-					}
-					Expect(clusterVMs).NotTo(BeEmpty(), "no guest VMs (masters/workers) to attach disk to")
-					baseKubeconfig = testClusterResources.BaseKubeconfig
+				if testClusterResources.BaseKubeconfig == nil {
+					Skip("VirtualDisk creation requires base cluster kubeconfig (Deckhouse virtualization). " +
+						"Set SSH_JUMP_HOST to the base cluster or use TEST_CLUSTER_CREATE_MODE=alwaysCreateNew.")
 				}
+				baseKubeconfig = testClusterResources.BaseKubeconfig
 				By("Selecting VirtualMachines in phase Running only (skip Migrating/Starting for stable disk attach)")
-				clusterVMs = e2eIntersectVMNamesRunning(e2eCtx, baseKubeconfig, ns, clusterVMs)
+				clusterVMs = e2eListClusterVMNames(e2eCtx, testClusterResources, ns)
 
 				numNodes := len(clusterVMs)
 				parallelism := 1 + rand.Intn(numNodes)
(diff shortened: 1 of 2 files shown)


e2e/tests/sds_node_configurator_test.go
```diff
@@ -487,6 +487,111 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
...
+
+				By("Step 3: Detaching and deleting the VirtualDisk to simulate device loss")
+				Expect(kubernetes.DetachAndDeleteVirtualDisk(
+					e2eCtx,
+					testClusterResources.BaseKubeconfig,
+					ns,
+					diskAttachment.AttachmentName,
+					diskAttachment.DiskName,
+				)).To(Succeed())
+				e2eDiskAttachments = nil
```
szhem: **Severity: Medium** — `e2eDiskAttachments = nil` wipes entire cleanup list

`e2eDiskAttachments` is a suite-level variable. Setting it to `nil` removes all entries, not just the one from this test. If other tests in the Ordered suite appended attachments before this test ran, those references are lost and teardown won't clean them up.

**Recommendation:** Remove only the specific attachment, or follow the pattern from the Size Reduction test (separate variables + `AfterEach`):

```go
for i, att := range e2eDiskAttachments {
    if att.DiskName == diskAttachment.DiskName {
        e2eDiskAttachments = append(e2eDiskAttachments[:i], e2eDiskAttachments[i+1:]...)
        break
    }
}
```

e2e/tests/sds_node_configurator_test.go
```diff
@@ -563,6 +668,345 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
...
+					NodeName:   realNodeName,
+					Consumable: true,
+					Path:       "/dev/e2e-nonexistent-device",
+					Size:       resource.MustParse("1Gi"),
+					Type:       "disk",
+					MachineID:  "e2e-fake-machine-id",
+				}
+				err = k8sClient.Update(e2eCtx, fakeBD)
+				if err != nil {
+					err = k8sClient.Status().Update(e2eCtx, fakeBD)
```
szhem, line 869: **Severity: Medium** — Status update via `Update()` instead of `Status().Update()` — wrong fallback order

In Kubernetes, `status` is a separate subresource. `Update()` updates spec/metadata but **does not persist status fields**. If `Update()` succeeds, the status changes are silently lost. The correct order is to try `Status().Update()` first, then fall back to full `Update()` only if the status subresource is not enabled.

Same issue appears on line 974.

**Recommendation:**

```go
err = k8sClient.Status().Update(e2eCtx, fakeBD)
if err != nil {
    err = k8sClient.Update(e2eCtx, fakeBD)
}
```

e2e/tests/sds_node_configurator_test.go
```diff
@@ -563,6 +668,345 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
...
+					var bd v1alpha1.BlockDevice
+					err := k8sClient.Get(e2eCtx, client.ObjectKey{Name: fakeBDName}, &bd)
+					g.Expect(apierrors.IsNotFound(err)).To(BeTrue(),
+						"fake BlockDevice %s should be deleted by the agent; current state: err=%v, consumable=%t, nodeName=%s",
+						fakeBDName, err, bd.Status.Consumable, bd.Status.NodeName)
+				}, 5*time.Minute, 10*time.Second).Should(Succeed())
+				By(fmt.Sprintf("Fake BlockDevice %s was deleted by the agent", fakeBDName))
+			})
+
+			It("Should revert manual modifications to an existing BlockDevice status", func() {
```
szhem, line 887: **Severity: Medium** — No cleanup for modified BlockDevice if test fails

This test modifies the `status.size` of a real, live BlockDevice to `999Ti`. If the 5-minute timeout expires (agent down, network issue, etc.), the test fails but the BD is left with a fake size. No `DeferCleanup` / `AfterEach` restores the original value, which can cascade into subsequent tests (e.g. Scheduler Extender) that rely on real BD sizes.

**Recommendation:** Add `DeferCleanup` to restore original size on failure:

```go
DeferCleanup(func() {
    var bd v1alpha1.BlockDevice
    if err := k8sClient.Get(e2eCtx, client.ObjectKey{Name: targetBD.Name}, &bd); err == nil {
        if !bd.Status.Size.Equal(originalSize) {
            bd.Status.Size = originalSize
            _ = k8sClient.Status().Update(e2eCtx, &bd)
        }
    }
})
```

e2e/tests/sds_node_configurator_test.go
```diff
@@ -563,6 +668,345 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
...
+					Skip("No BlockDevices in cluster to test modification revert")
+				}
+
+				var targetBD *v1alpha1.BlockDevice
+				for i := range bdList.Items {
+					bd := &bdList.Items[i]
+					if bd.Status.Path != "" && bd.Status.Size.Value() > 0 {
+						targetBD = bd
+						break
+					}
```
szhem, line 903: **Severity: Medium** — First available BD may be in use by an LVMVolumeGroup

This selects the first BD with a valid path and size, regardless of whether it's backing an LVG. Modifying its `status.size` to `999Ti` could trigger the LVG controller to react (reconciliation loop, error conditions), causing cascading failures during the test.

**Recommendation:** Prefer a BD that is `consumable` (not bound to a VG):

```go
if bd.Status.Path != "" && bd.Status.Size.Value() > 0 && bd.Status.Consumable {
    targetBD = bd
    break
}
```

e2e/tests/sds_node_configurator_test.go
```diff
@@ -487,6 +487,111 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
...
+							clusterVMs = append(clusterVMs, name)
+						}
+					}
+				}
+				if len(clusterVMs) == 0 {
+					vmNames, listErr := kubernetes.ListVirtualMachineNames(e2eCtx, testClusterResources.BaseKubeconfig, ns)
+					Expect(listErr).NotTo(HaveOccurred(), "list VirtualMachines on base cluster")
+					Expect(vmNames).NotTo(BeEmpty(), "no VirtualMachines in namespace %s on base cluster", ns)
+					clusterVMs = vmNames
+				}
```
szhem: **Severity: Low** — Duplicated VM list retrieval pattern

This 14-line block (get VMs from `VMResources`, fall back to `ListVirtualMachineNames`) is copied verbatim in the Size Reduction test (~line 720). Consider extracting into a helper like `getClusterWorkerVMs(ctx, resources, kubeconfig, ns) []string` to reduce duplication.

e2e/tests/sds_node_configurator_test.go
```diff
@@ -563,6 +668,345 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
...
+			It("Should delete a manually created BlockDevice that does not correspond to a real device", func() {
+				ensureSchedulerE2EK8sClient(testClusterResources, &k8sClient, e2eCtx)
+
+				By("Step 1: Getting a real node name from the cluster")
+				var nodeList corev1.NodeList
+				Expect(k8sClient.List(e2eCtx, &nodeList)).To(Succeed())
+				Expect(nodeList.Items).NotTo(BeEmpty(), "cluster must have at least one node")
+				realNodeName := nodeList.Items[0].Name
+
+				fakeBDName := e2eFakeBDPrefix + fmt.Sprintf("%d", rand.Intn(100000))
```
szhem: **Severity: Nit** — `fmt.Sprintf("%d", ...)` → `strconv.Itoa`

`strconv.Itoa(rand.Intn(100000))` is a more idiomatic way to convert an int to string.

szhem (commented):
## Review Summary

| Severity | Count |
|----------|-------|
| Critical | 0 |
| High | 0 |
| Medium | 4 |
| Low | 1 |
| Nit | 1 |

### Positive Observations

- Tests are well-structured with clear step-by-step `By()` annotations and informative assertion messages
- Graceful handling of webhook protection in the manual BD creation test (`IsForbidden`/`IsInvalid` → graceful skip/verify)
- `restartSDSNodeConfiguratorAgentOnNode` is a clean, reusable helper covering the full pod lifecycle (delete → gone → replacement running+ready)
- `AfterEach` cleanup in the Size Reduction test correctly strips finalizers before LVG deletion
- README documentation accurately describes each scenario and its assertions
- `Eventually` timeouts are appropriate for e2e (5-10 min with 10-15s polling)

e2e/tests/sds_node_configurator_test.go
kkozoriz: Duplicated at: 502–515, 725–738, and a close variant at 1344–1370

e2e/tests/sds_node_configurator_test.go
```diff
@@ -563,6 +673,361 @@ var _ = Describe("sds-node-configurator module e2e", Ordered, func() {
...
+						},
+						Type:  "Local",
+						Local: v1alpha1.LVMVolumeGroupLocalSpec{NodeName: nodeName},
+					},
+				}
+				Expect(k8sClient.Create(e2eCtx, lvg)).To(Succeed())
+
+				defer func() {
+					var current v1alpha1.LVMVolumeGroup
+					if err := k8sClient.Get(e2eCtx, client.ObjectKeyFromObject(lvg), &current); err == nil && current.Status.Phase != v1alpha1.PhaseReady {
```
kkozoriz: Maybe move this into the existing AfterEach? Feels weird to have a defer and an AfterEach for the same LVG side by side

kkozoriz (dismissed):

e2e/tests/sds_node_configurator_test.go
viktor-karpochev: BlockDevice has a status field, but I don't see a dedicated status subresource in the CRD, and the controller itself updates BlockDevice via regular Update() in images/agent/internal/controller/bd/discoverer.go, so I’d keep the current order here.

kkozoriz (approved):

PR #238, opened 2026-04-16
[agent] Add udev property parsers and DeviceMap event store

fastrapier:
## Summary

This is PR 1 of 3 in the effort to replace `lsblk` exec-based block device
discovery with an in-process netlink/udev/sysfs approach.

This PR introduces the foundational data layer (`internal/udev` package):

- **`Properties`** — typed struct for parsed udev environment, matching
  lsblk's field-resolution priority chains (serial, model, wwn).
- **`ParseProperties`** — parses raw `map[string]string` from netlink events
  into `Properties`. Returns error on invalid/missing MAJOR/MINOR, never
  returns partial data.
- **`DeviceMap`** — thread-safe concurrent store (`map[string]Properties`,
  keyed by `"major:minor"`). `HandleEvent` processes kernel uevents
  (add/change/remove/bind/unbind) and returns errors to the caller.
- **`ErrUnknownAction`** — sentinel error for programmatic handling of
  unexpected uevent actions.

## How this will be used

1. **PR 2** will add `FillFromCrawler` to `DeviceMap` for initial device
   population from sysfs crawler + udev DB enrichment.
2. **PR 3** will add sysfs readers, `BuildDevice` (Properties → internal.Device),
   `Snapshot`, and integrate everything into `scanner.Run` — replacing the
   current `lsblk` exec path.

## What is NOT in this PR

- No sysfs I/O (reading size, rotational, hotplug, slaves)
- No `BuildDevice` / `Snapshot` / scanner integration
- No changes to existing code — this is purely additive

## Test plan

- 33 unit tests covering property parsing, error handling, DeviceMap
  lifecycle (add/change/remove/bind/unbind/move/online/offline),
  concurrent access, and copy safety

### DIFF
--- a/images/agent/internal/udev/devicemap.go
+++ b/images/agent/internal/udev/devicemap.go
@@ -17,21 +17,24 @@ limitations under the License.
 package udev
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 )
+
+var ErrUnknownAction = errors.New("unknown action")
 
 // DeviceMap is a thread-safe in-memory store of block device properties
 // keyed by "major:minor". It is populated by netlink events (HandleEvent)
 // and provides a consistent snapshot via All().
 type DeviceMap struct {
 	mu      sync.RWMutex
-	devices map[string]UdevProperties
+	devices map[string]Properties
 }
 
 func NewDeviceMap() *DeviceMap {
 	return &DeviceMap{
-		devices: make(map[string]UdevProperties),
+		devices: make(map[string]Properties),
 	}
 }
 
@@ -39,7 +42,7 @@ func NewDeviceMap() *DeviceMap {
 // directly from the kernel (add, change, remove, bind, unbind, move, online,
 // offline). The env map is the raw udev environment from the event.
 func (dm *DeviceMap) HandleEvent(action string, env map[string]string) error {
-	props, err := ParseUdevProperties(env)
+	props, err := ParseProperties(env)
 	if err != nil {
 		return fmt.Errorf("handle event (action=%s, DEVNAME=%s): %w", action, env["DEVNAME"], err)
 	}
@@ -55,7 +58,7 @@ func (dm *DeviceMap) HandleEvent(action string, env map[string]string) error {
 	case "remove", "unbind", "offline":
 		delete(dm.devices, key)
 	default:
-		return fmt.Errorf("unknown action %q for device %s", action, key)
+		return fmt.Errorf("%w %q for device %s", ErrUnknownAction, action, key)
 	}
 
 	return nil
@@ -63,11 +66,11 @@ func (dm *DeviceMap) HandleEvent(action string, env map[string]string) error {
 
 // All returns a shallow copy of the device map. Callers may modify the
 // returned map without affecting the DeviceMap.
-func (dm *DeviceMap) All() map[string]UdevProperties {
+func (dm *DeviceMap) All() map[string]Properties {
 	dm.mu.RLock()
 	defer dm.mu.RUnlock()
 
-	cp := make(map[string]UdevProperties, len(dm.devices))
+	cp := make(map[string]Properties, len(dm.devices))
 	for k, v := range dm.devices {
 		cp[k] = v
 	}
--- a/images/agent/internal/udev/devicemap.go
+++ b/images/agent/internal/udev/devicemap.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2026 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package udev
+
+import (
+	"fmt"
+	"sync"
+)
+
+// DeviceMap is a thread-safe in-memory store of block device properties
+// keyed by "major:minor". It is populated by netlink events (HandleEvent)
+// and provides a consistent snapshot via All().
+type DeviceMap struct {
+	mu      sync.RWMutex
+	devices map[string]UdevProperties
+}
+
+func NewDeviceMap() *DeviceMap {
+	return &DeviceMap{
+		devices: make(map[string]UdevProperties),
+	}
+}
+
+// HandleEvent processes a single netlink uevent. The action string comes
+// directly from the kernel (add, change, remove, bind, unbind, move, online,
+// offline). The env map is the raw udev environment from the event.
+func (dm *DeviceMap) HandleEvent(action string, env map[string]string) error {
+	props, err := ParseUdevProperties(env)
+	if err != nil {
+		return fmt.Errorf("handle event (action=%s, DEVNAME=%s): %w", action, env["DEVNAME"], err)
+	}
+
+	key := DeviceKey(props.Major, props.Minor)
+
+	dm.mu.Lock()
+	defer dm.mu.Unlock()
+
+	switch action {
+	case "add", "change", "bind", "move", "online":
+		dm.devices[key] = props
+	case "remove", "unbind", "offline":
+		delete(dm.devices, key)
+	default:
+		return fmt.Errorf("unknown action %q for device %s", action, key)
+	}
+
+	return nil
+}
+
+// All returns a shallow copy of the device map. Callers may modify the
+// returned map without affecting the DeviceMap.
+func (dm *DeviceMap) All() map[string]UdevProperties {
+	dm.mu.RLock()
+	defer dm.mu.RUnlock()
+
+	cp := make(map[string]UdevProperties, len(dm.devices))
+	for k, v := range dm.devices {
+		cp[k] = v
+	}
+	return cp
+}
+
+// Len returns the number of devices currently tracked.
+func (dm *DeviceMap) Len() int {
+	dm.mu.RLock()
+	defer dm.mu.RUnlock()
+	return len(dm.devices)
+}
--- a/images/agent/internal/udev/devicemap.go
+++ b/images/agent/internal/udev/devicemap.go
@@ -22,6 +22,8 @@ import (
 	"sync"
 )
 
+// ErrUnknownAction is returned by HandleEvent when the uevent action
+// is not one of the recognized kernel actions.
 var ErrUnknownAction = errors.New("unknown action")
 
 // DeviceMap is a thread-safe in-memory store of block device properties
@@ -32,6 +34,7 @@ type DeviceMap struct {
 	devices map[string]Properties
 }
 
+// NewDeviceMap returns an empty, ready-to-use DeviceMap.
 func NewDeviceMap() *DeviceMap {
 	return &DeviceMap{
 		devices: make(map[string]Properties),
@@ -41,6 +44,13 @@ func NewDeviceMap() *DeviceMap {
 // HandleEvent processes a single netlink uevent. The action string comes
 // directly from the kernel (add, change, remove, bind, unbind, move, online,
 // offline). The env map is the raw udev environment from the event.
+//
+// Add-like actions (add, change, bind, move, online) insert or overwrite the
+// device entry. Remove-like actions (remove, unbind, offline) delete it.
+// This is intentional: we listen on the UdevEvent multicast group, so every
+// event carries a fully enriched property set, and a device in the offline or
+// unbound state is unavailable for I/O — there is no value in keeping it in
+// the map.
 func (dm *DeviceMap) HandleEvent(action string, env map[string]string) error {
 	props, err := ParseProperties(env)
 	if err != nil {
@@ -49,16 +59,17 @@ func (dm *DeviceMap) HandleEvent(action string, env map[string]string) error {
 
 	key := DeviceKey(props.Major, props.Minor)
 
-	dm.mu.Lock()
-	defer dm.mu.Unlock()
-
 	switch action {
 	case "add", "change", "bind", "move", "online":
+		dm.mu.Lock()
 		dm.devices[key] = props
+		dm.mu.Unlock()
 	case "remove", "unbind", "offline":
+		dm.mu.Lock()
 		delete(dm.devices, key)
+		dm.mu.Unlock()
 	default:
-		return fmt.Errorf("%w %q for device %s", ErrUnknownAction, action, key)
+		return fmt.Errorf("handle event: %w %q for device %s", ErrUnknownAction, action, key)
 	}
 
 	return nil
--- a/images/agent/internal/udev/devicemap_test.go
+++ b/images/agent/internal/udev/devicemap_test.go
@@ -214,3 +214,47 @@ func TestDeviceMap_ConcurrentAccess(t *testing.T) {
 	}
 	assert.Equal(t, 100, dm.Len())
 }
+
+func TestDeviceMap_ConcurrentReadWrite(t *testing.T) {
+	dm := NewDeviceMap()
+
+	var wg sync.WaitGroup
+	for i := 0; i < 50; i++ {
+		wg.Add(1)
+		go func(minor int) {
+			defer wg.Done()
+			env := makeEnv("8", fmt.Sprintf("%d", minor), fmt.Sprintf("sd%d", minor))
+			_ = dm.HandleEvent("add", env)
+		}(i)
+	}
+	for i := 0; i < 50; i++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			_ = dm.All()
+			_ = dm.Len()
+		}()
+	}
+	wg.Wait()
+}
+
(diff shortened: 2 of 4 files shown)


szhem (commented):
# Code Review

Clean, well-tested foundational layer. Findings below, grouped by severity.

## Summary

| Severity | Count |
|---|---|
| Critical | 0 |
| High | 1 |
| Medium | 3 |
| Low | 3 |
| Nit | 3 |

## Positive observations

- Correct error wrapping (`%w`) and sentinel errors (`ErrUnknownAction`) — aligns with Go Code Review Comments.
- `Properties` is a pure value type, so `All()` genuinely returns an unentangled snapshot.
- Parsing happens outside the lock in `HandleEvent` — minimal mutex hold time.
- `serial`/`model`/`wwn` priority chains faithfully mirror `lsblk get_properties_by_udev()` with source references in comments — easy to verify against util-linux.
- Tests cover every uevent action, parser priorities, `unhexmangle` edge cases (invalid hex, truncated escape), empty-map behavior, and copy safety of `All()`.
- `require.NoError` vs `assert.Equal` usage is consistent (fatal vs non-fatal).
- PR description clearly scopes this as step 1 of 3 — makes reviewing easier.

## PR-wide notes

- Consider adding a `Get(key) (Properties, bool)` helper in PR 2/3 — `All()` is the only read path today, which forces full-map copies for single-device lookups.
- No benchmark coverage — acceptable for foundational data layer, but worth revisiting once hot paths are known.

images/agent/internal/udev/devicemap.go
```diff
@@ -0,0 +1,85 @@
...
+	defer dm.mu.Unlock()
+
+	switch action {
+	case "add", "change", "bind", "move", "online":
+		dm.devices[key] = props
+	case "remove", "unbind", "offline":
+		delete(dm.devices, key)
+	default:
+		return fmt.Errorf("%w %q for device %s", ErrUnknownAction, action, key)
+	}
```
szhem, line 73: **Severity: High** — `offline` / `unbind` delete the entry; `bind` / `online` overwrite properties

`offline` on SCSI/transport is a transient state, not a physical removal. `unbind` only detaches the driver — the device still exists in sysfs. Deleting the entry drops serial/model/WWN that the subsequent `online`/`bind` event may not repopulate (the kernel only emits fields udev managed to re-read). If a consumer (the PR 3 scanner) calls `All()` between `offline` and `online`, the device disappears from the snapshot → spurious reconcile churn upstream (LVG/BD controllers).

Symmetrically, `bind` sits in the add-like branch and overwrites properties: if a device was already populated by a full `add` event, a later `bind` with a thin env will wipe fields back to empty.

**Recommendation:**
1. For `offline` / `unbind`, keep the entry and track state via flags in `Properties` (e.g. `Online bool`, `Bound bool`) — let the consumer decide how to treat a temporarily unavailable device.
2. For `change` / `bind` / `online` / `move`, merge: non-empty fields from the new env overwrite, empty fields leave existing values intact. This matches udevd behavior (it retains the last known set).
3. Alternatively, document the current policy explicitly in the `HandleEvent` docstring and coordinate with PR 2/3 so `Snapshot` distinguishes "removed" from "offline".

images/agent/internal/udev/devicemap.go
```diff
@@ -0,0 +1,85 @@
...
+
+package udev
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+)
+
+var ErrUnknownAction = errors.New("unknown action")
```
szhem, line 27: **Severity: Medium** — missing doc comment on exported symbol

Per Go Code Review Comments, every exported symbol should start with a doc comment beginning with the symbol name. `go doc` currently returns empty for `ErrUnknownAction`.

**Recommendation:**
```go
// ErrUnknownAction is returned by HandleEvent when the uevent action
// is not one of the recognized kernel actions.
var ErrUnknownAction = errors.New("unknown action")
```

images/agent/internal/udev/devicemap.go
```diff
@@ -0,0 +1,85 @@
...
+
+// DeviceMap is a thread-safe in-memory store of block device properties
+// keyed by "major:minor". It is populated by netlink events (HandleEvent)
+// and provides a consistent snapshot via All().
+type DeviceMap struct {
+	mu      sync.RWMutex
+	devices map[string]Properties
+}
+
+func NewDeviceMap() *DeviceMap {
```
szhem, line 38: **Severity: Medium** — missing doc comment on exported constructor

`DeviceMap`, `HandleEvent`, `All`, `Len` all have doc comments; `NewDeviceMap` does not. Inconsistent with the rest of the file and with Go style.

**Recommendation:**
```go
// NewDeviceMap returns an empty, ready-to-use DeviceMap.
func NewDeviceMap() *DeviceMap { ... }
```

images/agent/internal/udev/devicemap.go
```diff
@@ -0,0 +1,85 @@
...
+func (dm *DeviceMap) HandleEvent(action string, env map[string]string) error {
+	props, err := ParseProperties(env)
+	if err != nil {
+		return fmt.Errorf("handle event (action=%s, DEVNAME=%s): %w", action, env["DEVNAME"], err)
+	}
+
+	key := DeviceKey(props.Major, props.Minor)
+
+	dm.mu.Lock()
+	defer dm.mu.Unlock()
```
szhem: **Severity: Low** — `Lock` is held even for the unknown-action path

No shared state is touched on the `default` branch, so the mutex is unnecessary for that case. Cheap optimization, and it avoids queuing invalid events behind the lock under load.

**Recommendation:** validate `action` before `dm.mu.Lock()`:
```go
switch action {
case "add", "change", "bind", "move", "online",
     "remove", "unbind", "offline":
    // ok
default:
    return fmt.Errorf("%w %q for device %s", ErrUnknownAction, action, key)
}
dm.mu.Lock()
defer dm.mu.Unlock()
// ... mutating switch ...
```

images/agent/internal/udev/devicemap.go
```diff
@@ -0,0 +1,85 @@
...
+
+	switch action {
+	case "add", "change", "bind", "move", "online":
+		dm.devices[key] = props
+	case "remove", "unbind", "offline":
+		delete(dm.devices, key)
+	default:
+		return fmt.Errorf("%w %q for device %s", ErrUnknownAction, action, key)
```
szhem: **Severity: Nit** — error message word order

`fmt.Errorf("%w %q for device %s", ErrUnknownAction, action, key)` produces `unknown action "explode" for device 8:0`. Readable, but `%w` in the middle is unusual and inconsistent with the `handle event (...)` prefix used in the other error branch.

**Recommendation:**
```go
return fmt.Errorf("handle event: %w %q for device %s", ErrUnknownAction, action, key)
```

images/agent/internal/udev/devicemap.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	return nil
+}
+
+// All returns a shallow copy of the device map. Callers may modify the
+// returned map without affecting the DeviceMap.
+func (dm *DeviceMap) All() map[string]Properties {
```
szhem, line 80: **Severity: Nit** — `All()` name does not convey the copy contract

Readers have to consult the docstring to learn that the returned map is detached from internal state. Idiomatic alternatives in K8s-adjacent codebases: `Snapshot()` or `DeepCopyAll()`. `All()` paired with `Len()` reads like a raw accessor.

**Recommendation:** consider renaming to `Snapshot()` — self-documenting.

images/agent/internal/udev/properties.go
```diff
@@ -0,0 +1,141 @@
...
+
+package udev
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+type Properties struct {
```
szhem, line 27: **Severity: Medium** — missing doc comment on exported type

`Properties` is a central exported type but has no docstring. `go doc udev.Properties` will be empty.

**Recommendation:**
```go
// Properties is the parsed, normalized subset of a udev environment that
// matches lsblk's field-resolution rules for block device metadata.
type Properties struct { ... }
```

images/agent/internal/udev/properties.go
```diff
@@ -0,0 +1,141 @@
...
+	Model    string
+	WWN      string
+	FSType   string
+	PartUUID string
+	DMName   string
+	DMUUID   string
+	MDLevel  string
+}
+
+func ParseProperties(env map[string]string) (Properties, error) {
```
szhem, line 45: **Severity: Medium** — missing doc comment on exported function

`ParseProperties` needs a docstring describing the expected env contract (required MAJOR/MINOR, optional rest) and the error semantics ("never returns partial data" — stated in the PR description, but not in the code).

**Recommendation:**
```go
// ParseProperties parses a raw udev environment map into Properties.
// MAJOR and MINOR are required and must be non-negative integers;
// any parse error yields a zero Properties and a wrapped error.
func ParseProperties(env map[string]string) (Properties, error) { ... }
```

images/agent/internal/udev/properties.go
```diff
@@ -0,0 +1,141 @@
...
+	major, err := strconv.Atoi(env["MAJOR"])
+	if err != nil {
+		return Properties{}, fmt.Errorf("parse MAJOR: %w", err)
+	}
+
+	minor, err := strconv.Atoi(env["MINOR"])
+	if err != nil {
+		return Properties{}, fmt.Errorf("parse MINOR: %w", err)
+	}
```
szhem: **Severity: Medium** — `MAJOR`/`MINOR` accept negative integers

`strconv.Atoi("-5")` returns `-5, nil`. The kernel never emits negative device numbers, but there is no defensive validation, and `DeviceKey(-1, -2)` produces a well-formed key `"-1:-2"` that pollutes the map and will never match anything PR 2/3 lookups up.

**Recommendation:** use `strconv.ParseUint` with base 10 and bit size 32, or add an explicit non-negative check:
```go
major, err := strconv.ParseUint(env["MAJOR"], 10, 32)
if err != nil {
    return Properties{}, fmt.Errorf("parse MAJOR: %w", err)
}
```

images/agent/internal/udev/properties.go
```diff
@@ -0,0 +1,141 @@
...
+		return v
+	}
+	return env["ID_WWN"]
+}
+
+// normalizeWhitespace collapses whitespace runs and trims edges,
+// matching util-linux normalize_whitespace() used by lsblk after reading udev properties.
+func normalizeWhitespace(s string) string {
+	return strings.Join(strings.Fields(s), " ")
+}
```
szhem, line 103: **Severity: Low** — comment claims exact parity with `normalize_whitespace()` but implementation uses `strings.Fields`

`strings.Fields` splits on Unicode whitespace (`unicode.IsSpace`), while util-linux `normalize_whitespace()` splits on ASCII `isspace(3)`. In practice udev values are ASCII, so behavior matches, but the docstring asserts strict parity — either soften the wording or iterate byte-by-byte.

**Recommendation:** change the comment to `// normalizeWhitespace collapses whitespace runs and trims edges, approximating util-linux normalize_whitespace() for ASCII inputs.`

images/agent/internal/udev/properties.go
```diff
@@ -0,0 +1,141 @@
...
+
+func hexVal(c byte) int {
```
szhem: **Severity: Nit** — `hexVal` name is too generic for a helper private to `unhexmangle`

The function only exists to support `unhexmangle`. A more contextual name (`hexNibble`, `fromHexDigit`) makes intent obvious at the call site. Can also be inlined since `strconv.ParseUint(s[i+2:i+4], 16, 8)` covers the same job — see the related comment on `unhexmangle`.

images/agent/internal/udev/properties.go
```diff
@@ -0,0 +1,141 @@
...
+		if i+3 < len(s) && s[i] == '\\' && s[i+1] == 'x' &&
+			hexVal(s[i+2]) >= 0 && hexVal(s[i+3]) >= 0 {
+			b.WriteByte(byte(hexVal(s[i+2])<<4 | hexVal(s[i+3])))
+			i += 4
+			continue
+		}
+		b.WriteByte(s[i])
+		i++
+	}
+	return b.String()
```
szhem, line 121: **Severity: Low** — `unhexmangle` can be simplified via `strconv.ParseUint` and drop `hexVal` entirely

Current implementation rolls its own hex parsing. Equivalent, shorter version:
```go
func unhexmangle(s string) string {
    var b strings.Builder
    b.Grow(len(s))
    for i := 0; i < len(s); {
        if i+3 < len(s) && s[i] == '\\' && s[i+1] == 'x' {
            if v, err := strconv.ParseUint(s[i+2:i+4], 16, 8); err == nil {
                b.WriteByte(byte(v))
                i += 4
                continue
            }
        }
        b.WriteByte(s[i])
        i++
    }
    return b.String()
}
```
Deletes the `hexVal` helper and keeps identical semantics (verified against the existing test cases `\x20`, `\xZZ`, `\x4`).

images/agent/internal/udev/devicemap_test.go
```diff
@@ -0,0 +1,216 @@
...
+			errs[minor] = dm.HandleEvent("add", env)
+		}(i)
+	}
+	wg.Wait()
+
+	for i, err := range errs {
+		assert.NoError(t, err, "goroutine %d returned error", i)
+	}
+	assert.Equal(t, 100, dm.Len())
+}
```
szhem, line 216: **Severity: Medium** — concurrency test exercises only writers

100 goroutines all call `HandleEvent("add", ...)`. No reader runs concurrently, so `go test -race` would not surface a hypothetical read/write race in `All()` or `Len()`. Also no interleaving of `add` + `remove` for the same key.

**Recommendation:** add a second test where half the goroutines write and half call `All()` / `Len()` in a loop, plus a test where pairs of goroutines fight over the same key with alternating `add` / `remove`.

images/agent/internal/udev/devicemap.go
fastrapier: The listener connects via `conn.Connect(netlink.UdevEvent)` — this is group 2 (udev multicast), not raw kernel events. Every event, including `bind`, arrives with the full udevd-enriched property set, so the "thin env overwrite" scenario doesn't apply here.

For `offline`/`unbind` deleting entries: the device is genuinely unavailable for I/O in these states. Consumers (BD/LVG controllers) can't use an offline device, so keeping it in the map has no practical benefit and would require the consumer to filter it out anyway.

Documented this policy in the `HandleEvent` docstring — added a paragraph explaining why add-like actions overwrite and remove-like actions delete, and why this is intentional given the UdevEvent socket.

images/agent/internal/udev/devicemap.go
fastrapier: Fixed.

images/agent/internal/udev/devicemap.go
fastrapier: Fixed.

images/agent/internal/udev/devicemap.go
fastrapier: Fixed. Went a step further — instead of a double switch, each branch acquires the mutex only around its own mutation. The `default` path returns before touching any shared state.

images/agent/internal/udev/devicemap.go
fastrapier: Fixed.

images/agent/internal/udev/devicemap.go
fastrapier: PR 3 will introduce a `Snapshot` method on `DeviceMap` that returns `[]internal.Device` (built from Properties + sysfs data). Renaming `All()` to `Snapshot()` now would create a naming collision with a different return type. Keeping `All()` for the raw `map[string]Properties` accessor.

images/agent/internal/udev/properties.go
fastrapier: Fixed.

images/agent/internal/udev/properties.go
fastrapier: Fixed.

images/agent/internal/udev/properties.go
fastrapier: Fixed. Switched to `strconv.ParseUint(_, 10, 32)` for both MAJOR and MINOR.

images/agent/internal/udev/properties.go
fastrapier: Fixed.

images/agent/internal/udev/properties.go
fastrapier: Resolved by replacing the entire `unhexmangle` implementation with `strconv.ParseUint` — `hexVal` is removed entirely.

images/agent/internal/udev/properties.go
fastrapier: Fixed. `hexVal` removed.

images/agent/internal/udev/devicemap_test.go
fastrapier: Fixed. Added two tests:
- `TestDeviceMap_ConcurrentReadWrite` — 50 writers + 50 goroutines calling `All()`/`Len()` concurrently
- `TestDeviceMap_ConcurrentAddRemoveSameKey` — 100 goroutines alternating `add`/`remove` on the same key

kneumoin (approved):
NOTE: scanner integration/parity tests.

PR #268, opened 2026-05-13
chore(ci): use check_previous_channel_release action

duckhawk:
## Summary
- Replaces the per-repo `.github/check_previous_channel_release.sh` with the new shared composite action [`deckhouse/modules-actions/check_previous_channel_release@v14`](https://github.com/deckhouse/modules-actions/pull/73).
- The shared action fixes the previous-channel `version.json` parser so it handles both the single-line `{"version":"v0.1.14"}` layout and the pretty-printed `{ "version": "v0.1.0" }` (multi-line) output produced by `yq -o=json`.
- No functional change to the deploy flow; the matrix of editions and the order of steps stay the same.

## Test plan
- [ ] Trigger a non-alpha channel deploy (e.g. promote to `beta`) once `deckhouse/modules-actions#73` is merged into `v14`.
- [ ] Verify the `Check previous channel release for <module>` step passes when versions match and fails (rejecting the promotion) when they don't.


duckhawk:
Pushed a follow-up commit that switches the `registry` input for `check_previous_channel_release` from `vars.PROD_REGISTRY` to `vars.PROD_REGISTRY_READ`. The composite action now swaps the domain portion of `module_source` with that `registry` input before pulling, so the previous-channel image can be read from the read-only registry mirror without changing the deploy target. Depends on [deckhouse/modules-actions#75](https://github.com/deckhouse/modules-actions/pull/75) being merged into `v14` first.

mrwesadex-flant (approved):

PR #286, opened 2026-06-02
Feat/legacy bd name saver

fastrapier:
This pull request enhances the block device discovery logic to better handle legacy non-consumable block devices whose unique names may change due to hardware or metadata updates. It introduces a mechanism to match new candidates with existing legacy block devices based on stable attributes (like serial number or WWN), ensuring continuity and preventing unnecessary device recreation. Comprehensive tests have been added to validate these changes and cover various edge cases.

Block device reconciliation improvements:
* Added logic in `blockDeviceReconcile` to detect and adopt legacy non-consumable block devices when a new candidate's generated name differs but stable identifiers match, preventing redundant device removal and recreation.
* Introduced the `findLegacyNonConsumableBlockDevice` function and supporting matching logic (`legacyNonConsumableBlockDeviceMatches`, `sameBlockDeviceTypeAndSize`) to robustly identify legacy non-consumable devices based on node, consumable status, and stable hardware attributes.

Testing enhancements:
* Added a dedicated test suite for `findLegacyNonConsumableBlockDevice`, covering matching by node, consumable status, type, size, serial, ambiguous matches, and edge cases to ensure correct legacy device adoption.
* Added a new test to `BlockDeviceReconcile` to verify that existing non-consumable block devices are updated (not recreated) when only the generated name changes due to hardware property modifications.

General:
* Updated copyright years in `discoverer.go` and `discoverer_test.go`. [[1]](diffhunk://#diff-a8709c961a667ab569b5b9d40eab05f22da36e23d92bb9c520ca0cde7948814cL2-R2) [[2]](diffhunk://#diff-79cad4228611f153baf328f823bab74b2b5dcbb98c8a5567dc75bf5ef951269bL2-R2)

### DIFF
--- a/images/agent/internal/controller/bd/discoverer.go
+++ b/images/agent/internal/controller/bd/discoverer.go
@@ -674,6 +674,7 @@ func legacyNonConsumableBlockDeviceMatches(
 
 	return candidate.Path != "" &&
 		candidate.Path == blockDevice.Status.Path &&
+		candidate.Type == blockDevice.Status.Type &&
 		candidate.Size.Value() == blockDevice.Status.Size.Value()
 }
 
--- a/images/agent/internal/controller/bd/discoverer.go
+++ b/images/agent/internal/controller/bd/discoverer.go
@@ -144,6 +144,13 @@ func (d *Discoverer) blockDeviceReconcile(ctx context.Context) (bool, error) {
 		if !exist {
 			legacyBlockDevice, found := findLegacyNonConsumableBlockDevice(candidate, apiBlockDevices)
 			if found {
+				d.log.Info(fmt.Sprintf(
+					`[RunBlockDeviceController] found legacy non-consumable BlockDevice, candidate name: "%s", legacy name: "%s", path: "%s"`,
+					candidate.Name, legacyBlockDevice.Name, candidate.Path,
+				))
+				// Adopt the legacy name only for this update path. The candidate
+				// slice keeps the new name, and old non-consumable devices are not
+				// removed by removeDeprecatedAPIDevices.
 				candidate.Name = legacyBlockDevice.Name
 				blockDevice = legacyBlockDevice
 				exist = true
@@ -665,16 +672,29 @@ func legacyNonConsumableBlockDeviceMatches(
 	if candidate.PVUuid != "" && candidate.PVUuid == blockDevice.Status.PVUuid {
 		return true
 	}
-	if candidate.VGUuid != "" && candidate.VGUuid == blockDevice.Status.VGUuid {
-		return true
-	}
 	if candidate.PartUUID != "" && candidate.PartUUID == blockDevice.Status.PartUUID {
 		return true
 	}
 
+	if sameBlockDeviceTypeAndSize(candidate, blockDevice) {
+		if wwn := candidate.GetWWN(); wwn != "" {
+			return wwn == blockDevice.Status.Wwn
+		}
+		if serial := candidate.GetSerial(); serial != "" {
+			return serial == blockDevice.Status.Serial
+		}
+		if blockDevice.Status.Wwn != "" || blockDevice.Status.Serial != "" {
+			return false
+		}
+	}
+
 	return candidate.Path != "" &&
 		candidate.Path == blockDevice.Status.Path &&
-		candidate.Type == blockDevice.Status.Type &&
+		sameBlockDeviceTypeAndSize(candidate, blockDevice)
+}
+
+func sameBlockDeviceTypeAndSize(candidate internal.BlockDeviceCandidate, blockDevice v1alpha1.BlockDevice) bool {
+	return candidate.Type == blockDevice.Status.Type &&
 		candidate.Size.Value() == blockDevice.Status.Size.Value()
 }
 
--- a/images/agent/internal/controller/bd/discoverer.go
+++ b/images/agent/internal/controller/bd/discoverer.go
@@ -1,17 +1,17 @@
 /*
-Copyright 2025 Flant JSC
+	Copyright 2026 Flant JSC
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+		http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
 */
 
 package bd
@@ -141,6 +141,15 @@ func (d *Discoverer) blockDeviceReconcile(ctx context.Context) (bool, error) {
 	// create new API devices
 	for _, candidate := range candidates {
 		blockDevice, exist := apiBlockDevices[candidate.Name]
+		if !exist {
+			legacyBlockDevice, found := findLegacyNonConsumableBlockDevice(candidate, apiBlockDevices)
+			if found {
+				candidate.Name = legacyBlockDevice.Name
+				blockDevice = legacyBlockDevice
+				exist = true
+			}
+		}
+
 		if exist {
 			addToDeleteListIfNotMatched := func(blockDevice v1alpha1.BlockDevice) {
 				if !deviceMatchesSelector(&blockDevice) {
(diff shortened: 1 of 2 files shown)


images/agent/internal/controller/bd/discoverer.go
```diff
@@ -621,6 +630,54 @@ func isBlockDeviceDeprecated(blockDevice string, actualCandidates map[string]str
...
+	if blockDevice.Status.NodeName != candidate.NodeName || blockDevice.Status.Consumable {
+		return false
+	}
+
+	if candidate.PVUuid != "" && candidate.PVUuid == blockDevice.Status.PVUuid {
+		return true
+	}
+	if candidate.VGUuid != "" && candidate.VGUuid == blockDevice.Status.VGUuid {
+		return true
+	}
```
fastrapier: **Severity: High** — Matching by `VGUuid` can bind a candidate to a different physical device

**Tier: Contract must** — LVM data model: one VG contains many PVs that all share the same `vgUUID`.

`VGUuid` identifies a *volume group*, not a device. A VG can contain several PVs; each PV is a distinct block device with its own `BlockDevice` CR, and they all carry the same `vgUUID` (see `candidate.VGUuid = pv.VGUuid` in `getBlockDeviceCandidates`).

The `PVUuid` check above correctly fails for the pair `candidate PV-A` vs `CR of PV-B` (`A != B`), but this next check `VGUuid == VGUuid` then returns `true`, because PV-A and PV-B belong to the same VG. The candidate is matched against the CR of a *different* physical device.

**Failure mode.** If only the CR of PV-B remains in the cluster (PV-A's CR was previously deleted) and a candidate for PV-A arrives with a new hash name, `findLegacyNonConsumableBlockDevice` finds exactly one match — PV-B's CR — and `blockDeviceReconcile` overwrites PV-B's `status` with PV-A's data. The CR now describes the wrong device and PV-B loses its representation. In the common "both CRs exist" case the ambiguity guard saves you (2 matches → `false`), but at the cost of skipping the migration and re-creating a duplicate.

**Recommendation:** `VGUuid` is not a device identifier and must not be used as a standalone equality criterion. Drop this branch (PVUuid + PartUUID + path/type/size already cover real device identity), or, if a VGUuid-based match is genuinely needed, pair it with a distinguishing field (`Path`/`Serial`) and cover it with a dedicated unit test.

```go
if candidate.PVUuid != "" && candidate.PVUuid == blockDevice.Status.PVUuid {
	return true
}
// VGUuid alone does not distinguish PVs within the same VG — do not match on it.
if candidate.PartUUID != "" && candidate.PartUUID == blockDevice.Status.PartUUID {
	return true
}
```

images/agent/internal/controller/bd/discoverer.go
```diff
@@ -621,6 +630,54 @@ func isBlockDeviceDeprecated(blockDevice string, actualCandidates map[string]str
...
+	if candidate.PartUUID != "" && candidate.PartUUID == blockDevice.Status.PartUUID {
+		return true
+	}
+
+	return candidate.Path != "" &&
+		candidate.Path == blockDevice.Status.Path &&
+		candidate.Type == blockDevice.Status.Type &&
+		candidate.Size.Value() == blockDevice.Status.Size.Value()
```
fastrapier, line 698: **Severity: Medium** — `Path`-based fallback match is not stable across reboots / re-plug

**Tier: Strong recommendation** — device paths in Linux are not persistent (`/dev/sdX` can move). See https://wiki.archlinux.org/title/Persistent_block_device_naming

This fallback applies to non-consumable devices without `PVUuid`/`VGUuid`/`PartUUID` (e.g. a hot-plug disk, or a disk with a filesystem directly). `Path` (`/dev/sdz`, `/dev/sdb`) is not persistent: after a reboot or re-plug the same path can be occupied by a different physical device.

**Failure mode.** `apiBlockDevices` may still hold a stale CR with `Status.Path == /dev/sdz` from a device that no longer exists. A new physical device of the same `Type` and `Size` that now occupies `/dev/sdz` will be treated as the legacy device and "update" the wrong CR — effectively re-identifying the device. This is exactly the class of devices for which the hash name is already unstable (no stable identifiers), so the risk is real.

**Recommendation:** this is a defensible trade-off (better than a duplicate), but narrow it and document it. At minimum add a comment that the `Path` branch is best-effort and only valid when no stable identifier exists. Better: also compare `Serial`/`Wwn` when the candidate has them, so two different devices with matching `Path+Type+Size` are not conflated. Add a negative unit test for "Path+Type+Size match but different Serial → no match".

images/agent/internal/controller/bd/discoverer.go
```diff
@@ -141,6 +141,15 @@ func (d *Discoverer) blockDeviceReconcile(ctx context.Context) (bool, error) {
...
 	for _, candidate := range candidates {
 		blockDevice, exist := apiBlockDevices[candidate.Name]
+		if !exist {
+			legacyBlockDevice, found := findLegacyNonConsumableBlockDevice(candidate, apiBlockDevices)
+			if found {
+				candidate.Name = legacyBlockDevice.Name
+				blockDevice = legacyBlockDevice
+				exist = true
+			}
+		}
```
fastrapier, line 158: **Severity: Medium** — Legacy match and ambiguity are silent (no logs, no metrics)

**Tier: Operational must** — controller-runtime structured-logging practice; every create/update/delete in this file is already logged (see the `updated/created/deleted APIBlockDevice` log lines).

A non-trivial event happens here: a candidate with a new hash name has its name swapped to a legacy one. The later log prints `updated APIBlockDevice, name: <old name>`, and an operator cannot tell why the device lives under a "stale" name.

Symmetrically, on ambiguity `findLegacyNonConsumableBlockDevice` silently returns `false` (the `if found { return ..., false }` branch inside the loop), and reconcile creates a duplicate — the very duplicate this PR exists to remove — with no trace in logs or metrics.

**Failure mode.** When duplicates appear in production there is no way to distinguish "legacy match failed due to ambiguity" from other causes. The drift between "found legacy" and "created new" is unobservable.

**Recommendation:** add a `Debug` log at the match point (old name → matched legacy name, and which identifier matched) and a `Warning` log on ambiguity listing the candidate names. Optionally add a counter (e.g. `legacy_match_total{result="matched|ambiguous"}`) next to the existing reconcile metrics.

images/agent/internal/controller/bd/discoverer_test.go
```diff
@@ -501,6 +553,132 @@ func TestBlockDeviceCtrl(t *testing.T) {
 		assert.Equal(t, len(deviceName[4:]), 40, "device name does not contains sha1 sum")
 	})
 
+	t.Run("FindLegacyNonConsumableBlockDevice", func(t *testing.T) {
```
fastrapier, line 556: **Severity: Medium** — The `PVUuid` / `VGUuid` / `PartUUID` matching branches are untested

**Tier: Strong recommendation** — Google "Testing on the Toilet": new behavior should be covered by unit tests; the PR description itself states "comprehensive unit tests".

These tests cover: a `Path+Size` match on the same node, rejection for a consumable candidate, rejection on different `Type`, and rejection on ambiguity. But three of the four branches of `legacyNonConsumableBlockDeviceMatches` — match by `PVUuid`, by `VGUuid`, by `PartUUID` — have no case at all. The `VGUuid` branch is precisely where the High-severity defect lives, and a test would expose it.

**Recommendation:** add table cases:
- (a) match by `PVUuid`;
- (b) match by `PartUUID`;
- (c) **negative** for `VGUuid` — candidate PV-A and a CR for PV-B with the same `VGUuid` but different `PVUuid`, expecting no match (this should fail on the current code and pass after the `VGUuid` fix);
- (d) `PVUuid` taking priority over `Path`.

images/agent/internal/controller/bd/discoverer.go
```diff
@@ -1,17 +1,17 @@
...
-See the License for the specific language governing permissions and
-limitations under the License.
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
 */
 
 package bd
```
fastrapier, line 17: **Severity: Low** — License-header reformatting (tabs + year bump) desyncs files

**Tier: Nit / personal preference** — consistency with the rest of the repository.

Leading tabs were added before every line of the Apache license text (the license text is conventionally left verbatim), and the year was changed `2025 → 2026`. Sibling files in the same package (e.g. `internal/block_device_candidate.go`) keep the original tab-free format and year `2025`, so this change introduces an inconsistency within one package. `gofmt` does not touch block-comment contents, which is why the linter passed.

**Recommendation:** revert the header reformatting and the year change where there is no substantive edit; if a year range is intended, use a single consistent form (`Copyright 2025-2026`) across all touched files.

images/agent/internal/controller/bd/discoverer.go
```diff
@@ -141,6 +141,15 @@ func (d *Discoverer) blockDeviceReconcile(ctx context.Context) (bool, error) {
 	// create new API devices
 	for _, candidate := range candidates {
 		blockDevice, exist := apiBlockDevices[candidate.Name]
+		if !exist {
+			legacyBlockDevice, found := findLegacyNonConsumableBlockDevice(candidate, apiBlockDevices)
+			if found {
+				candidate.Name = legacyBlockDevice.Name
```
fastrapier, line 154: **Severity: Nit** — Local mutation of `candidate.Name` inside the `range` loop deserves a note

**Tier: Nit / personal preference** — readability.

`candidate` is a copy of the slice element, so `candidate.Name = legacyBlockDevice.Name` is local and does **not** affect `actualCandidates` in `removeDeprecatedAPIDevices` (which uses the original new name). This is correct (the legacy non-consumable device is not deleted, since `shouldDeleteBlockDevice` only removes consumable devices), but it is non-obvious.

**Recommendation:** a one-line comment such as "name is swapped only for this in-place update; actualCandidates keeps the new name — the old one is not deleted because it is non-consumable" would save a future reader the analysis.

fastrapier (commented):
## Code Review — legacy non-consumable BlockDevice matching

The overall approach is sound: name is `sha1(NodeName+Wwn+Model+Serial+PartUUID)`, so when `Model`/`Serial`/`Wwn` change the hash name changes, the old non-consumable `BlockDevice` is not garbage-collected (`shouldDeleteBlockDevice` only cleans consumable devices), and a duplicate appears. Re-using the legacy name to update in place is the right fix. Findings below are about correctness of the matching predicate, fallback stability, observability, and test coverage.

### Summary

| Severity | Count |
|---|---|
| Critical | 0 |
| High | 1 |
| Medium | 3 |
| Low | 1 |
| Nit | 1 |

Main blocker before merge: the **High** finding — matching by `VGUuid` can overwrite the CR of a different physical device that belongs to the same volume group.

### Positive observations

- The ambiguity guard (`found` → a second match returns `false`) is implemented correctly and makes the result deterministic regardless of Go's randomized map iteration order.
- The early return for a `Consumable` candidate plus the symmetric `!blockDevice.Status.Consumable` check correctly scope the feature to legacy non-consumable devices only.
- Functions are small and single-purpose; the matching predicate is cleanly extracted into `legacyNonConsumableBlockDeviceMatches`.
- Good integration test `updates_existing_non_consumable_block_device_when_new_hash_name_differs`: it reproduces the real scenario (changed `Model` → different hash name) and asserts both the absence of a duplicate and preservation of the old name.

CI is green (Go tests, linter, coverage, CVE scan, Gitleaks, DCO all pass).

images/agent/internal/controller/bd/discoverer.go
fastrapier: fixed

duckhawk (approved):

PR #310, opened 2026-07-03
test(e2e): migrate suite onto the storage-e2e dvp provider scheme

duckhawk:
## Description

Migrates the sds-node-configurator e2e suite onto storage-e2e's **clusterprovider (dvp)** scheme — the same split bootstrap → run-tests flow sds-object uses with commander — instead of the bespoke `alwaysCreateNew`/`alwaysUseExisting` bootstrap plus local storage-e2e source patching.

- **go.mod**: bump storage-e2e to `main` (carries the dvp provider + `Connect`/`ConnectBase`). A **temporary** `replace => ../../storage-e2e` builds against the local checkout carrying the dvp `Connect` change until deckhouse/storage-e2e#44 merges; in CI `e2e-run-tests.sh` overrides it with the checked-out ref. Re-pin `require` and drop the replace once that PR is in `main`.
- **Suite**: `e2eEnsureSharedNestedTestCluster` prefers the provider path when `E2E_TEST_CLUSTER_PROVIDER` is set (`cluster.CreateOrConnectToTestCluster` → dvp `Connect` for the nested API + `ConnectBase` for the base cluster, surfaced as `res.BaseKubeconfig` for VirtualDisk ops). Legacy modes kept for local runs.
- **cfg**: drop hard `required` tags; validate conditionally — provider runs don't need `KUBE_CONFIG_PATH`/`CREATE_MODE`, legacy runs still require the full SSH/StorageClass set.
- **node-SSH**: fall back to `E2E_DVP_BASE_CLUSTER_SSH_*` / `E2E_DVP_VM_SSH_USER` and accept a raw-PEM key, so the lsblk/VG specs can reach worker VMs under dvp.
- Add `tests/cluster_config.ci.yml` (dvp; sds-node-configurator `modulePullOverride: ${E2E_MODULE_IMAGE_TAG}`) and `.github/workflows/e2e-tests.yml` (reusable storage-e2e pipeline, `cluster_provider: dvp`, label gate `e2e/dvp/run`).
- Drop `patches/` and the Makefile `fix-mod-permissions` hack — the new scheme no longer patches the storage-e2e module cache.

> **Depends on** deckhouse/storage-e2e#44 (dvp `Connect`/`ConnectBase`). Merge that first, then drop the temporary `replace` here and re-pin `go.mod`.

The existing suite pinned an old storage-e2e revision and patched its source at build time (`fix-mod-permissions`: `sed` on timeouts + a local `.diff`), which is brittle and drifts from upstream. It also predates storage-e2e's clusterprovider abstraction, so it couldn't use the shared, CI-integrated bootstrap/teardown pipeline. This aligns sds-node-configurator with the sds-object e2e scheme (thin caller → reusable `storage-e2e/.github/workflows/e2e.yml`) while keeping every existing spec, and moves cluster bring-up/module-enable into the maintained dvp provider.

- Adding the `e2e/dvp/run` label on a PR runs the reusable pipeline: bootstrap (provision VMs on the DVP base cluster, `dhctl` bootstrap, join nodes, enable modules from `cluster_config.ci.yml`) → run-tests (suite connects through the dvp provider) → teardown.
- `go build ./...` and test-compile pass. All BlockDevice/LVMVolumeGroup/scheduler/netlink/stability specs are preserved (labels unchanged: smoke = `!stress-test`).
- Local runs via the legacy `TEST_CLUSTER_CREATE_MODE` path keep working.

## Checklist
- [ ] e2e tests passed. 


duckhawk:
Superseded by #321 (`test/e2e-refactor-iter1`), which landed the storage-e2e dvp migration on main with a different layout (one file per scenario + `e2e/framework` package, thin reusable-workflow CI caller). Closing to avoid a second, divergent e2e stack.

PR #321, opened 2026-07-22
Refactor and enhance E2E testing framework with utilities

fastrapier:
This pull request introduces several significant refactorings and new utilities to the e2e test framework, focusing on simplifying configuration, improving testability, and adding new helpers for block device and virtualization testing. The main changes include removing unused configuration options and validation, adding a new stress test configuration loader, and introducing new helper functions for working with block devices, disks, LVM, and remote execution. It also adds comprehensive tests for the new configuration logic.

### Configuration Simplification and Refactoring

* Removed unused fields (`SSH`, `KubeConfigPath`, `DKPLicenceKey`, `LogLevel`, `YamlClusterConfig`, `CreateMode`, `Cleanup`) and related validation logic from `Config` and `TestCluster`, significantly simplifying configuration for e2e tests. (`e2e/cfg/config.go`, `e2e/cfg/validation.go`, `e2e/cfg/mode.go`) [[1]](diffhunk://#diff-55166303164c6707dd2d74a3b15178c36a49e795cf6f99055a054e507bc81711L25-R30) [[2]](diffhunk://#diff-55166303164c6707dd2d74a3b15178c36a49e795cf6f99055a054e507bc81711L62-L66) [[3]](diffhunk://#diff-07bf5066e51bc2b245d317bcf84d2a3077ffe07a4b77346ca7356785969a0140L1-L25) [[4]](diffhunk://#diff-67dc9389f1d2f77c33947901d40c70255d3459f2b2a30b317020611e04186799L1-L36)
* Updated tests in `config_test.go` to match the new configuration structure and added tests for the new stress configuration loader. (`e2e/cfg/config_test.go`)

### New Configuration Loader for Stress Tests

* Added a new `Stress` struct and `LoadStress` function to parse and validate environment variables for stress tests, including clamping and normalization logic. (`e2e/cfg/stress.go`)

### New Framework Utilities

* Added helper functions for working with block devices, including deterministic device naming, waiting for new devices, and triggering LVM discovery. (`e2e/framework/blockdevice.go`)
* Added a utility to resize VirtualDisks via the Kubernetes dynamic client. (`e2e/framework/disk.go`)
* Added a function for executing commands on nodes with error handling. (`e2e/framework/exec.go`)
* Added LVM helper functions for parsing command output and generating teardown scripts. (`e2e/framework/lvm.go`)
* Added a parser for `lsblk` output to extract device information. (`e2e/framework/parse.go`)

### Documentation

* Clarified in the `e2e/README.md` that e2e tests must be run serially due to shared cluster state. (`e2e/README.md`)


duckhawk (approved):

e2e/framework/parse.go
```diff
@@ -0,0 +1,72 @@
...
+	Size      string
+	SizeBytes int64
+}
+
+// ParseLsblk parses `lsblk -b -P -o NAME,SIZE,SERIAL,PATH` output (KEY="value"
+// per line), keyed by PATH.
+//
+// TODO: split-by-space breaks on values containing spaces (e.g. disk model).
+// Root fix = upstream lsblk parser in storage-e2e.
+func ParseLsblk(out string) map[string]LsblkLine {
```
szhem, line 37: **Severity: High**

**Tier: Strong recommendation — project convention (this PR adds `cfg/config_test.go`) + Google Testing on the Toilet**

The new `framework` package adds pure / near-pure logic — `ParseLsblk` here, plus `Poll` (`poll.go`), `CountPVsInVG` / `VGInListing` / `ThinPoolDataLVPresent` (`lvm.go`) and `BlockDeviceName` (`blockdevice.go`) — but ships no unit tests. `find e2e -name '*_test.go'` shows tests only under `cfg/` and `tests/`; the `framework/` directory has none.

This is not a "tests would be nice" note: this PR *establishes* the unit-test convention for helper logic by adding `cfg/config_test.go` for `LoadStress`. The framework helpers are the same shape of code and carry more edge cases, yet ship untested.

**Failure mode:** a regression in `ParseLsblk` field-splitting or in `Poll`'s `(done, err)` timeout contract surfaces only as a flaky / wrong-reason failure deep inside a 3.5h e2e run, where it is blamed on the cluster rather than the helper. The whole point of extracting these into `framework` was to make them testable in isolation.

**Recommendation:** add `framework/parse_test.go`, `framework/lvm_test.go`, `framework/poll_test.go`. Highest-value cases:
- `Poll`: (a) `done=true,err=nil` → `nil`; (b) `done=true,err!=nil` → returns that err immediately, no second call (use a call counter); (c) never done → timeout error wraps the last non-nil cond error; (d) ctx cancelled mid-wait → `ctx.Err()`; (e) zero timeout still runs cond once.
- `ParseLsblk`: happy path, empty input, missing `PATH` (row skipped), non-numeric `SIZE` (`SizeBytes==0`), space-in-`SERIAL`.
- `ThinPoolDataLVPresent`: matches `pool` and `pool_tdata`, ignores non-`t` `lv_attr`, tolerates the `[...]`-bracketed internal-LV form.
- `BlockDeviceName`: golden test pinning the `dev-<sha1>` output against a known agent value, since the doc comment requires it stay in sync with the agent's `createUniqDeviceName`.

e2e/cfg/stress.go
```diff
@@ -0,0 +1,84 @@
...
+	if s.BatchSize > s.Target {
+		return nil, fmt.Errorf("stress config: BatchSize (%d) exceeds Target (%d)", s.BatchSize, s.Target)
+	}
+
+	if s.MaxVMBlockDevices > maxVMBlockDevicesCeiling {
+		s.MaxVMBlockDevices = maxVMBlockDevicesCeiling
+	}
+	if s.Target > s.MaxVMBlockDevices {
+		s.Target = s.MaxVMBlockDevices
+	}
```
szhem, line 68: **Severity: Medium**

**Tier: Operational must — internal inconsistency between the stated intent and the code**

The comment on `maxVMBlockDevicesCeiling` (line 26) says the cap is 16 total *"leaving one slot reserved for the boot device"*, which implies at most **15** attachable *data* disks — and the default `MaxVMBlockDevices` is correctly `15`, consistent with that reasoning. But this clamp allows the value back up to `16`, and `Target` is then clamped to that same `16`.

Verified against the [Deckhouse Virtualization Platform requirements](https://deckhouse.io/products/virtualization-platform/documentation/about/requirements.html): *"maximum number of block devices to be attached is 16"* — that 16 is the **total**, and the VM's boot disk consumes one of the 16, so the usable-for-data ceiling really is 15, exactly as the comment says and the code contradicts.

**Failure mode:** with an override such as `E2E_STRESS_MAX_VM_BLOCK_DEVICES=16` (or any value ≥16, which clamps to 16) and `E2E_STRESS_MAX_VG_STRICT=true`, the ramp targets 16 data disks. Together with the boot disk that is 17 attachments against a hard cap of 16, so the 16th `AttachDisk` fails and strict mode asserts `readyTotal == target` → a spurious test failure attributed to the driver, not to this off-by-one. The default path (15) is safe; only overrides hit this.

**Recommendation:**
```go
// vmBlockDeviceHardCap is the total VMBDA-per-VM limit in Deckhouse virtualization.
// One slot is the boot device, so at most vmBlockDeviceHardCap-1 data disks attach.
const vmBlockDeviceHardCap = 16
const maxDataBlockDevicesCeiling = vmBlockDeviceHardCap - 1 // 15
...
if s.MaxVMBlockDevices > maxDataBlockDevicesCeiling {
    s.MaxVMBlockDevices = maxDataBlockDevicesCeiling
}
```
and update `TestLoadStressMaxVMClampedToCeiling` to expect `15`. If the full 16 (including boot) is really intended, fix the comment instead — but the strict-mode math argues for 15.

e2e/cfg/config_test.go
```diff
@@ -22,44 +22,109 @@ import (
...
+		"TEST_CLUSTER_NAMESPACE",
+		"E2E_DVP_BASE_CLUSTER_STORAGE_CLASS",
+		"MODULES_MODULE_TAG",
+	} {
+		assert.NoError(t, os.Unsetenv(k))
+	}
 
-	err = os.Setenv("DKP_LICENSE_KEY", "test-key")
-	assert.NoError(t, err, "Failed to set env DKP_LICENSE_KEY")
+	assert.NoError(t, os.Setenv("TEST_CLUSTER_NAMESPACE", "custom-ns"))
```
szhem, line 56: **Severity: Low**

**Tier: Strong recommendation — [`testing.T.Setenv`](https://pkg.go.dev/testing#T.Setenv) is the idiomatic auto-restoring form**

`TestLoad` and `TestLoadOverrides` set env via `os.Setenv` / `os.Unsetenv` and never restore, while the sibling `TestLoadStress*` tests correctly use `t.Setenv` (which registers a `Cleanup` to restore the prior value). These two tests leave `TEST_CLUSTER_NAMESPACE=custom-ns`, `MODULES_MODULE_TAG=pr-123`, etc. set for whatever runs after them in the package.

**Failure mode:** order-dependent flakiness — a future test added to `package cfg` that reads these vars would observe leaked values. Today it is masked only because each `TestLoad*` re-`Unsetenv`s at its own start.

**Recommendation:** replace the manual `os.Setenv` / `os.Unsetenv` with `t.Setenv(...)` for the keys each test sets and drop the `Unsetenv` loop; `t.Setenv` both sets and schedules restoration. (Note it is incompatible with `t.Parallel()`, which these tests correctly do not use.)

e2e/framework/blockdevice.go
```diff
@@ -0,0 +1,86 @@
...
+// BlockDevice that is present now but was not in `before` (set-diff by Name). It errors if, at timeout,
+// there are zero new devices, and errors immediately if more than one new device appears.
+func WaitNewConsumableBlockDevice(ctx context.Context, restCfg *rest.Config, node string, before []kubernetes.BlockDevice, timeout time.Duration) (kubernetes.BlockDevice, error) {
+	beforeSet := make(map[string]struct{}, len(before))
+	for _, bd := range before {
+		beforeSet[bd.Name] = struct{}{}
+	}
+
+	var result kubernetes.BlockDevice
+	err := Poll(ctx, 5*time.Second, timeout, func(ctx context.Context) (bool, error) {
```
szhem, line 49: **Severity: Nit**

`Poll(ctx, 5*time.Second, timeout, ...)` embeds the poll interval as a literal. It is used once and is self-explanatory, so this is optional — but since the sibling timeouts (`stressBDDiscoveryTimeout` etc.) are named constants, a `bdDiscoveryPollInterval = 5 * time.Second` would keep the family consistent. Feel free to ignore.

PR #330, opened 2026-08-04
feat(scheduler-extender): take Pod PVCs from the extra-pvcs annotation

fastrapier:
## What

Teach `sds-common-scheduler-extender` to read the PVCs a Pod needs from the
`scheduler.deckhouse.io/extra-pvcs` annotation, as a **second source in addition to**
`pod.Spec.Volumes`.

- **Key:** `scheduler.deckhouse.io/extra-pvcs`, one constant (`pkg/consts/consts.go`).
- **Value:** CSV of PVC names — `"pvc-a,pvc-b"`. Entries are trimmed, empties ignored.
- **Namespace:** always the Pod's own.

The collection point is `getManagedPVCsFromPod` (`pkg/scheduler/func.go`): it now walks a
deduplicated union of the spec-volume claim names and the CSV names from the annotation.
Once a PVC is in `managedPVCs`, the existing pipeline (`getStorageClassesUsedByPVCs` →
`extractRequestedSize` → local/replicated split → LVG/space filter → `createReservations`)
treats it exactly like a spec-sourced one. Both handlers (`filter`, `prioritize`) inherit
it, and the 60s reservation is created automatically.

### The hint contract

An annotation is a **best-effort hint, not a mount** — the Pod runs fine without that
volume — so an unresolvable name must never cost the Pod its nodes. A spec-sourced PVC
keeps the opposite rule: the Pod genuinely cannot run without it, so failing to resolve one
still fails the whole scheduling request.

Carrying that distinction past the collection point needs the PVC's origin, so
`getManagedPVCsFromPod` returns a second map marking the annotation-only names, and it is
honoured at two places:

1. **The PVC itself is missing.** `IsNotFound` on the `Get` skips the name (it may
   legitimately not exist yet); the skipped names are reported in one aggregated `Warning`.
2. **The PVC exists but cannot be resolved** — no `lvm-type` parameter, a deleted RSC/RSP,
   a PV gone from under a bound PVC. `extractRequestedSize` reports those names in a new
   `dropped` return value instead of failing; `filter`/`prioritize` then remove them from
   `managedPVCs`, exactly as they already do for `dropPVCsWithMissingSC`, and short-circuit
   if nothing is left. The Pod degrades to its pre-annotation behavior.

Not yet covered by (2): errors from `discoverProvisionerForPVC` and `getNewControlPlane`
inside `getManagedPVCsFromPod`, and from `getStorageClassesUsedByPVCs` / node lookups in
the handlers, still fail the request for an annotation-only PVC. See Known gaps.

### Behavior for Pods without the annotation

Unchanged, with **one deliberate exception**: `requestForPVC` now clamps the bound-PVC
delta at zero.

```go
RequestedSize: max(requested-pv.Spec.Capacity.Storage().Value(), 0)
```

LVM rounds an LV up to the extent boundary, so a PV backing a decimal-unit request ("10G")
is larger than the request itself and the delta went negative. `Cache.GetReservedSpace`
sums reservation sizes without checking the sign and `getAvailableSpace` computes
`baseFree - reserved`, so a negative reservation **inflated** the pool's free space for
every other Pod scheduled in the same 60s window. This is a fix on the main (spec-sourced)
path, not on the annotation path, and it applies whether or not the annotation is present.

Everything else for an annotation-free Pod is byte-identical, pinned by tests including the
pre-existing `TestShouldProcessPod` regression suite.

Kaiten US 67033244. A DVP VM with a hotplug disk is two Pods:

- `virt-launcher` (`d8v-vm-*`) runs the VM, and KubeVirt **deliberately** keeps hotplug
  volumes out of its `spec.volumes`;
- the hotplug attachment Pod (`d8v-hp-*`) carries the PVC and is pinned to the launcher's
  node by `nodeAffinity`.

So the launcher schedules **first and blind** — free space and replica proximity are never
consulted — and the attachment Pod is then forced onto whatever node the launcher picked.
If the volume can't live there, the attachment Pod hangs `Pending` while the launcher is
already on the wrong node.

Confirmed live on the `base` DVP stand (2026-08-03): launcher `d8v-vm-worker-1-2nzxw` has
only an `emptyDir` in `spec.volumes` and no PVC at all; its attachment Pod
`d8v-hp-worker-1-kph6b` holds the PVC on the same node. The extender is registered without
`managedResources`, so kube-scheduler *does* call `filter` for the launcher — the hook
exists, there was just nothing for it to act on.

**This is a best-effort hint, not a guarantee.** The extender is registered with
`ignorable: true` (if it times out, kube-scheduler schedules without it), the reservation
lives 60s and is not renewed, and only *hotplug-at-VM-start* is covered — hotplugging into
an already-running VM is explicitly out of scope, since `filter` for that launcher will
never run again.

- A Pod **without** the annotation: no observable change, except the negative-reservation
  clamp described above.
- A `virt-launcher`-shaped Pod (no PVC in `spec.volumes`) **with** the annotation: the
  extender stops being a no-op for it — it filters out nodes whose LVMVolumeGroup lacks
  space and places a 60s reservation on the survivors, so the launcher lands where the
  hotplug volume can actually be provisioned.
- An annotation naming a PVC of a foreign provisioner, or one that doesn't exist, or one
  the extender cannot resolve, or a malformed value: ignored with a `Warning`, and the Pod
  is scheduled as it would have been before this annotation existed.

Known mis-annotation risk: within a namespace, any Pod can name any PVC, and for
`Local`-access replicated PVCs that means a **hard reject** of nodes, not merely a bad
placement. Blast radius is namespace-bounded; validating who may set the annotation is
deliberately deferred until a real abuse case appears.

## Checklist

- [ ] **e2e tests passed** — no e2e in this PR. The annotation e2e specs are deliberately
      split out: they need their own design pass and land separately. This PR is the
      scheduler-extender change plus its unit and handler-level coverage.
- [ ] Documentation updated according to the changes — **no user-facing docs on purpose:**
      the final annotation key is still pending agreement with the other consumers, so
      publishing it now would document an unagreed contract.
- [ ] Changes were tested in the Kubernetes cluster manually — not yet.

## Test coverage

**Parsing and collection** (`pkg/scheduler/extra_pvcs_test.go`):

- CSV parser edge cases; the annotation key is pinned by a test, since it's a cross-module
  contract.
- `getManagedPVCsFromPod` table: no annotation (backward-compat), annotation-only, several
  names, spec+annotation dedup (a duplicate counts as spec-sourced), missing annotation PVC
  skipped, foreign provisioner filtered, malformed value, and **missing spec PVC still
  errors**. Each case also asserts which names come back marked annotation-only.

**Hint downgrade** (`pkg/scheduler/func_test.go`):

- `TestExtractRequestedSize_HintOnlyDowngrade` — every failure point of `requestForPVC`
  from both sides: RSC missing, RSP missing, no `lvm-type` parameter, PV gone from under a
  bound PVC; each asserted as *dropped* for an annotation-only PVC and as *still failing*
  for a spec-sourced one.
- `extractRequestedSize` must not mutate the caller's PVC map — the dropped names come back
  as a return value.
- The negative-delta clamp: a bound PVC whose PV is larger than the request reserves 0.

**Handler level** (`pkg/scheduler/extra_pvcs_test.go`), driving the real HTTP handlers:

- `/filter` local space-reject with the reservation actually created;
- `/filter` no-annotation no-op;
- `/filter` with an unbound **replicated** PVC at `volumeAccess: Local`;
- `/filter` with an unresolvable annotation PVC — all nodes survive and nothing is reserved;
- `/prioritize` scores nodes for an annotation PVC and narrows the reservation to the final
  node list.

**e2e** — none in this PR, on purpose (see the checklist above).

## Known gaps

- **No e2e coverage in this PR at all.** Behavior is pinned by unit and handler-level tests
  only. Note also that the e2e stand (`e2e/tests/cluster_config.yml`) installs only
  `sds-local-volume`, so whenever the e2e specs do land, the replicated path still cannot be
  covered there without adding `sds-replicated-volume` to the stand — today that path is
  covered at handler level by `TestFilter_AnnotationPVC_ReplicatedLocalAccess` and at unit
  level by the RSC/RSP cases of `TestExtractRequestedSize_HintOnlyDowngrade`.
- **The hint downgrade is not applied everywhere.** An annotation-only PVC still fails the
  whole scheduling request if `discoverProvisionerForPVC` or `getNewControlPlane` errors
  inside `getManagedPVCsFromPod`, or if `getStorageClassesUsedByPVCs` / the node lookups in
  the handlers error. Most of those are transient and kube-scheduler retries; the one that
  does not self-heal is a bound annotation PVC whose PV was deleted and whose StorageClass
  is also gone, which sends `discoverProvisionerForPVC` into a PV lookup that fails
  permanently.
- **The reservation for an annotation PVC never narrows to a single node.** `filter` spreads
  it over every surviving node and `prioritize` narrows it to the final node list, but the
  last step — narrowing to one node once VolumeBinding stamps `selected-node` on the PVC —
  never happens, because the Pod does not mount the PVC and no binding cycle touches it. So
  the reservation stays spread across all candidate nodes until the 60s TTL expires,
  temporarily understating free space on each of them for other Pods.
- **`local` + already-bound annotation PVC is not filtered meaningfully** — for a bound PVC
  the requested delta is ≈0, so the space check passes trivially and nothing verifies the
  node actually holds the LV. This is status quo, not a regression (before this change the
  extender returned all nodes for such a Pod), but it is worth knowing before anyone relies
  on annotation PVCs for bound local volumes.
- The `managedResources`-is-empty precondition lives in control-plane-manager, outside this
  repo, and can drift independently of this code.

### DIFF
--- a/images/sds-common-scheduler-extender/pkg/consts/consts.go
+++ b/images/sds-common-scheduler-extender/pkg/consts/consts.go
@@ -23,6 +23,14 @@ const (
 	LvmTypeParamKey         = "local.csi.storage.deckhouse.io/lvm-type"
 	LVMVolumeGroupsParamKey = "local.csi.storage.deckhouse.io/lvm-volume-groups"
 
+	// PodExtraPVCsAnnotation lists, as a comma-separated value, the names of
+	// PersistentVolumeClaims a Pod needs taken into account when scheduling but
+	// does not mount itself (e.g. a KubeVirt hotplug disk, which is attached by a
+	// separate attachment Pod pinned to the launcher's node). The PVCs are always
+	// looked up in the Pod's own namespace. It is a best-effort scheduling hint:
+	// unknown or foreign-provisioner names are ignored.
+	PodExtraPVCsAnnotation = "scheduler.deckhouse.io/extra-pvcs"
+
 	Thick = "Thick"
 	Thin  = "Thin"
 )
--- a/images/sds-common-scheduler-extender/pkg/consts/consts.go
+++ b/images/sds-common-scheduler-extender/pkg/consts/consts.go
@@ -1,17 +1,17 @@
 /*
-	Copyright 2026 Flant JSC
+Copyright 2025 Flant JSC
 
-	Licensed under the Apache License, Version 2.0 (the "License");
-	you may not use this file except in compliance with the License.
-	You may obtain a copy of the License at
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
 
-		http://www.apache.org/licenses/LICENSE-2.0
+    http://www.apache.org/licenses/LICENSE-2.0
 
-	Unless required by applicable law or agreed to in writing, software
-	distributed under the License is distributed on an "AS IS" BASIS,
-	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-	See the License for the specific language governing permissions and
-	limitations under the License.
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
 */
 
 package consts
--- a/images/sds-common-scheduler-extender/pkg/consts/consts.go
+++ b/images/sds-common-scheduler-extender/pkg/consts/consts.go
@@ -1,17 +1,17 @@
 /*
-Copyright 2025 Flant JSC
+	Copyright 2026 Flant JSC
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+	Licensed under the Apache License, Version 2.0 (the "License");
+	you may not use this file except in compliance with the License.
+	You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+		http://www.apache.org/licenses/LICENSE-2.0
 
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
+	Unless required by applicable law or agreed to in writing, software
+	distributed under the License is distributed on an "AS IS" BASIS,
+	WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+	See the License for the specific language governing permissions and
+	limitations under the License.
 */
 
 package consts
@@ -29,6 +29,12 @@ const (
 	// separate attachment Pod pinned to the launcher's node). The PVCs are always
 	// looked up in the Pod's own namespace. It is a best-effort scheduling hint:
 	// unknown or foreign-provisioner names are ignored.
+	//
+	// Contract for the producer: every PVC named here MUST already exist by the
+	// time the Pod is created. The extender resolves the names once, while running
+	// filter/prioritize for this Pod; a name that does not resolve is skipped and
+	// there is no second attempt — the hint is silently lost and the Pod is
+	// scheduled as blindly as it was before this annotation existed.
 	PodExtraPVCsAnnotation = "scheduler.deckhouse.io/extra-pvcs"
 
 	Thick = "Thick"
--- a/images/sds-common-scheduler-extender/pkg/scheduler/extra_pvcs_test.go
+++ b/images/sds-common-scheduler-extender/pkg/scheduler/extra_pvcs_test.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2025 Flant JSC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package scheduler
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/deckhouse/sds-node-configurator/images/sds-common-scheduler-extender/pkg/consts"
+)
+
+func TestPodExtraPVCsAnnotationKey(t *testing.T) {
+	// The key is a cross-module contract (producer lives in the virtualization
+	// module). Pin the exact string so a rename cannot happen silently.
+	assert.Equal(t, "scheduler.deckhouse.io/extra-pvcs", consts.PodExtraPVCsAnnotation)
+}
+
+func TestParseExtraPVCNames(t *testing.T) {
+	tt := []struct {
+		name  string
+		value string
+		want  []string
+	}{
+		{name: "empty value", value: "", want: nil},
+		{name: "single name", value: "pvc-a", want: []string{"pvc-a"}},
+		{name: "two names", value: "pvc-a,pvc-b", want: []string{"pvc-a", "pvc-b"}},
+		{name: "surrounding spaces are trimmed", value: " pvc-a , pvc-b ", want: []string{"pvc-a", "pvc-b"}},
+		{name: "empty entries are ignored", value: ",,pvc-a,,pvc-b,", want: []string{"pvc-a", "pvc-b"}},
+		{name: "only separators and spaces", value: " , , ", want: nil},
+		{name: "order is preserved", value: "b,a,c", want: []string{"b", "a", "c"}},
+		{name: "duplicates are kept (dedup happens later)", value: "pvc-a,pvc-a", want: []string{"pvc-a", "pvc-a"}},
+	}
+
+	for _, tc := range tt {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.want, parseExtraPVCNames(tc.value))
+		})
+	}
+}
--- a/images/sds-common-scheduler-extender/pkg/scheduler/extra_pvcs_test.go
+++ b/images/sds-common-scheduler-extender/pkg/scheduler/extra_pvcs_test.go
@@ -17,19 +17,27 @@ limitations under the License.
 package scheduler
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"sort"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	d8commonapi "github.com/deckhouse/sds-common-lib/api/v1alpha1"
 	"github.com/deckhouse/sds-node-configurator/images/sds-common-scheduler-extender/pkg/consts"
 	"github.com/deckhouse/sds-node-configurator/images/sds-common-scheduler-extender/pkg/logger"
+	srv "github.com/deckhouse/sds-replicated-volume/api/v1alpha1"
 )
 
 func TestPodExtraPVCsAnnotationKey(t *testing.T) {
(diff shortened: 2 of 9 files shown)


fastrapier (commented):
## Code review — extra-pvcs annotation

Reviewed at commit `beebd0b`. Verified locally on that tree: `go build -tags=ee ./...` clean, `go test -tags=ee -race ./pkg/scheduler/...` green (3.2 s).

Two findings below were confirmed with throwaway probe tests whose output is quoted verbatim; the probe file was deleted afterwards, the tree is unchanged.

| Severity | Count |
|---|---|
| Critical | 0 |
| High | 2 |
| Medium | 4 |
| Low | 2 |
| Nit | 1 |

The only blocking one is the first inline comment on `func.go` — it cancels the property this PR claims ("scheduling never fails because of it") and turns an annotation typo into a VM that never starts. The three findings below sit on code outside the diff, so they are here rather than inline.

---

### 3. Negative reservation size for a bound annotation PVC inflates free space for every other Pod

**Severity: Medium**

**Tier: Operational must — `getAvailableSpace` subtracts `reserved` from free space; a negative addend causes LVG overcommit**

**Locations:**

- [`pkg/scheduler/func.go:712-721`](https://github.com/deckhouse/sds-node-configurator/blob/beebd0b526ae959f6334ca0f7c8ed40a37e1d9e0/images/sds-common-scheduler-extender/pkg/scheduler/func.go#L712-L721)
- [`pkg/scheduler/filter.go:455-459`](https://github.com/deckhouse/sds-node-configurator/blob/beebd0b526ae959f6334ca0f7c8ed40a37e1d9e0/images/sds-common-scheduler-extender/pkg/scheduler/filter.go#L455-L459)

For a bound PVC, `extractRequestedSize` computes `requestedSize = requests.storage - pv.capacity`. LVM allocates an LV in whole extents (4 MiB by default), so a request in decimal units (`size: 10G`, common in DVP) yields a PV **larger** than the request and the difference goes negative. Neither `extractRequestedSize` nor `createReservations` clamps it, and `Cache.GetReservedSpace` simply sums the sizes of active reservations. In `getAvailableSpace` this is `available := baseFree - reserved`, so a negative `reserved` **increases** the pool's apparent free space — for every Pod scheduled in the same window, not just the one that caused it.

The subtraction itself predates this PR, but this PR opens a new path to it: before the change, a launcher Pod with a hotplug disk created no reservation at all (`managedPVCs` was empty and `filter` took the early return); now it does. This is exactly the scenario the description calls out as "`local` + already-bound annotation PVC is not filtered meaningfully" — but the consequence is stronger than stated there: it is not only that the space check passes trivially, it is that a negative reserve lands in the cache.

**Failure mode**, probe with request 10 GiB and PV 11 GiB, annotation pointing at that PVC:

```
reservation size=-1073741824 pools=[lvg-a]
GetReservedSpace(lvg-a) = -1073741824
```

Note that the normal cleanup does not kick in here: `RemoveReservation` for a bound PVC is called from the PVC watcher's `UpdateFunc`, and a bound PVC that stops changing emits no `Update` event. The reservation lives out the full 60 s TTL and is refreshed on every `filter` call while the launcher is being scheduled.

**Recommendation:** clamp at the single place where the value is computed — this covers spec PVCs too:

```go
// func.go, extractRequestedSize
requestedSize = pvc.Spec.Resources.Requests.Storage().Value() - pv.Spec.Capacity.Storage().Value()
if requestedSize < 0 {
    // PV rounded up to the LVM extent boundary: nothing left to reserve.
    requestedSize = 0
}
```

Plus a table test on `extractRequestedSize` with `pv.capacity > requests.storage` expecting `RequestedSize == 0`.

---

### 4. An annotation-only PVC's reservation is never narrowed to the chosen node

**Severity: Medium**

**Tier: Operational must — over-reservation across all candidate nodes degrades scheduling decisions for unrelated Pods**

**Locations:**

- [`pkg/controller/pvc_watcher_cache.go:111-119`](https://github.com/deckhouse/sds-node-configurator/blob/beebd0b526ae959f6334ca0f7c8ed40a37e1d9e0/images/sds-common-scheduler-extender/pkg/controller/pvc_watcher_cache.go#L111-L119)
- [`pkg/scheduler/filter.go:426-459`](https://github.com/deckhouse/sds-node-configurator/blob/beebd0b526ae959f6334ca0f7c8ed40a37e1d9e0/images/sds-common-scheduler-extender/pkg/scheduler/filter.go#L426-L459)

`createReservations` places the reservation on the pools of **all** surviving nodes — that is by design, narrowing happens later. But both narrowing mechanisms are tied to the PVC actually being mounted by the Pod being scheduled:

- `narrowReservationsToFinalNodes` in `prioritize` narrows to the final node list, not to a single node;
- the PVC watcher narrows to a concrete node based on the `volume.kubernetes.io/selected-node` annotation, which kube-scheduler's `VolumeBinding` plugin sets **only** for PVCs referenced from `pod.spec.volumes`.

For an annotation-only PVC, binding the launcher does **not** produce `selected-node`, so there is nothing to narrow and no trigger to do it: the reservation for the full volume size stays spread across every candidate node for the entire 60 s TTL.

```mermaid
flowchart TD
    F["filter: launcher Pod<br>PVC only in the annotation"] --> R["createReservations:<br>reserve 10 GiB on pools of<br>node-a, node-b, node-c"]
    R --> P["prioritize:<br>narrowReservationsToFinalNodes<br>narrows to the final node list"]
    P --> B["kube-scheduler binds launcher to node-a"]
    B --> S{"did volume.kubernetes.io/selected-node<br>appear on the PVC?"}
    S -- "No: PVC is not in spec.volumes" --> STUCK["reservation stays on a+b+c<br>for the full 60 s TTL"]
    S -- "Yes: ordinary spec PVC" --> NARROW["PVC watcher narrows to one node"]
```

**Failure mode:** while a single VM is being scheduled, N-1 nodes look 10 GiB (or whatever was requested) poorer than they are. On a batch VM start this systematically skews decisions and in the worst case produces false space rejections. It is bounded by the TTL, so no data loss — but it is also not "the same behaviour a spec PVC gets", which is how the description reads.

**Recommendation:** narrow the hint PVC's reservation to the node the Pod landed on from the existing hook — `prioritize` already knows the final node list, and the attachment Pod will overwrite the reservation later anyway. The minimal variant is a shorter TTL specifically for annotation-only PVCs, since they never take part in a binding cycle and waiting for `selected-node` from them is pointless:

```go
// filter.go, createReservations
ttl := defaultReservationTTL
if hintOnly[pvc.Name] {
    // No VolumeBinding cycle will ever narrow this reservation: the Pod does
    // not mount the PVC, so selected-node never appears. Keep it short.
    ttl = hintReservationTTL // e.g. 15s
}
schedulerCache.AddReservation(pvcKey, ttl, pvcReq.RequestedSize, poolKeys)
```

At the very least, document this in the `createReservations` comment so the next reader does not assume narrowing works here as well.

---

### 6. No e2e coverage for the new user-visible behaviour

**Severity: Medium**

**Tier: Strong recommendation — `e2e/README.md` plus the existing spec set in `e2e/tests/` as the established practice in this repo for user-observable changes**

The feature changes observable scheduler behaviour for a real user-facing object (a VM Pod with a hotplug disk), yet it is covered only at unit and handler level against a fake client. Nothing verifies that it changes anything in a live cluster; the "Changes were tested in the Kubernetes cluster manually" checkbox is unchecked too.

Separately: the reason for deferring e2e is stated explicitly in the description and looks sound (the specs need their own design pass; the stand in `e2e/tests/cluster_config.yml` installs only `sds-local-volume`, so the replicated path cannot be covered there anyway). That is a deliberate call by the author and I am not arguing against it here — recording it as an open item so it does not get lost, not as an objection to merging.

**Recommendation:** leave a link in the PR (or in the linked card) to the task carrying the e2e spec and its minimal content, so it does not stay "for later" without an address. Minimal scenario: a Pod with empty `spec.volumes` and the annotation pointing at a pending PVC of a local SC whose LVG exists on only one of the nodes → the Pod lands on that node; the same Pod without the annotation → lands anywhere. A separate item for adding `sds-replicated-volume` to the stand is worth filing too, otherwise the replicated path stays without e2e forever.

---

### Positive observations

- **The change is anchored in the right place.** The whole feature is confined to one PVC-collection point (`getManagedPVCsFromPod`); `filter`, `prioritize`, reservations and the local/replicated split are untouched. This is precisely the case where a new capability does not smear across layers — the opposite of [Shotgun Surgery](https://martinfowler.com/bliki/ShotgunSurgery.html).
- **A real logger bug is fixed along the way.** Replacing `log = log.WithValues("PVC", pvcName)` with a local `pvcLog` stops `PVC` values from accumulating on the shared logger across iterations — before this, log lines for the second PVC carried both names. The fix is not mentioned in the description, but it is valuable.
- **Backward compatibility is pinned by tests, not by prose.** `TestPodPVCNames` ("no annotation returns only spec volumes"), the "missing spec PVC still fails the request (unchanged behavior)" case and the handler-level `TestFilter_NoAnnotation_PodWithoutPVCsIsNoOp` together cover exactly the property that is most expensive to break silently.
- **`fromSpec` as an origin marker is a clean solution.** The "a name in both sources appears once and counts as spec-sourced" case, with its explicit assert message, shows the source precedence was reasoned about rather than arrived at by accident.
- **The annotation key is pinned by its own test** (`TestPodExtraPVCsAnnotationKey`) with a stated reason: it is a cross-module contract and a silent rename must fail. For a string another repository reads, that is the right guard.
- **An honest PR description.** The "Known gaps" section lists real limitations (no e2e, pointless filtering for bound local PVCs, `managedResources` living outside this repo) rather than only the strengths. Two of my findings — 3 and 6 — grew out of items the author found first, which measurably shortened this review.
- **Turning `if volume.PersistentVolumeClaim != nil { ... }` into a `continue` guard clause** removed a nesting level from the main loop, in line with [Effective Go](https://go.dev/doc/effective_go) on linear control flow.

images/sds-common-scheduler-extender/pkg/scheduler/func.go
```diff
@@ -422,50 +423,128 @@ func getNodeNames(inputData ExtenderArgs) ([]string, error) {
...
+			// single aggregated Warning after the loop (see below) instead of one
+			// Warning per PVC, so a stale/mistyped annotation on a Pod that gets
+			// retried by kube-scheduler doesn't flood the logs at the default
+			// (INFO) log level.
+			if apierrors.IsNotFound(err) && !fromSpec[pvcName] {
+				pvcLog.Debug(fmt.Sprintf("[getManagedPVCsFromPod] PVC listed in the %s annotation does not exist, skipping it", consts.PodExtraPVCsAnnotation))
+				missingAnnotationPVCs = append(missingAnnotationPVCs, pvcName)
 				continue
 			}
+			return nil, fmt.Errorf("[getManagedPVCsFromPod] error getting PVC: %v", err)
```
fastrapier: **Severity: High** — An error resolving a hint PVC turns into a rejection of all nodes, leaving the launcher Pod `Pending` forever

**Tier: Contract must — the contract this PR states itself ("An annotation naming a PVC of a foreign provisioner, or one that doesn't exist, or a malformed value: silently ignored, scheduling never fails because of it")**

The careful "a hint must never break scheduling" guard is implemented **only** here, and only for one case — `IsNotFound` on the PVC `Get` itself. As soon as a PVC makes it into `managedPVCs`, the fact that it came from the annotation is lost: `fromSpec` is not propagated any further. Any error downstream becomes fatal for the whole request again.

The nearest such break is [`extractRequestedSize`](https://github.com/deckhouse/sds-node-configurator/blob/beebd0b526ae959f6334ca0f7c8ed40a37e1d9e0/images/sds-common-scheduler-extender/pkg/scheduler/func.go#L688-L730): it returns an error if `deviceType` cannot be determined for a PVC, if a replicated SC has no RSC/RSP, or if a bound PVC's PV is gone. In [`filter`](https://github.com/deckhouse/sds-node-configurator/blob/beebd0b526ae959f6334ca0f7c8ed40a37e1d9e0/images/sds-common-scheduler-extender/pkg/scheduler/filter.go#L129-L135) that error goes to `writeFailAllNodesResponse` — a **valid HTTP 200 with a zero-length node list**. `ignorable: true` does not help here: it covers timeouts and 5xx, not a well-formed "no suitable nodes" answer.

**Failure mode:** a Pod that does not even mount the volume becomes unschedulable. For DVP this means the VM never starts — a regression from the pre-PR behaviour, where such a launcher scheduled fine (blindly, but it scheduled).

Confirmed with a probe test using a local SC without the `local.csi.storage.deckhouse.io/lvm-type` parameter, with the annotation pointing at a PVC of that SC:

```
NodeNames=[] FailedNodes=map[node-a:unable to extract request size:
  [extractRequestedSize] unable to determine device type for PVC default/pvc-hotplug]
```

Before this PR the same Pod would have got `[node-a]` back.

A missing `lvm-type` is not the only way into that branch, just the shortest reproducer. More realistic ones: an RSC deleted while the SC survives, an RSP lagging behind the cluster, a bound PVC whose PV was removed. The general shape of the problem is not a specific parameter — it is that past the boundary of this function the hint becomes a hard requirement again.

Note the asymmetry: in `prioritize` the same error is returned as `http.Error(500)`, which with `ignorable: true` genuinely is ignored. So the "a hint does not break scheduling" contract holds in `prioritize` and is violated in `filter` — exactly where it costs the Pod.

**Recommendation:** propagate the PVC's origin to the point where a decision is made about it, and for an annotation-only PVC degrade to "drop this PVC" rather than "reject all nodes". The cheapest form is to return the set of hint names from `getManagedPVCsFromPod` and consult it when `extractRequestedSize` fails:

```go
// func.go
func getManagedPVCsFromPod(...) (map[string]*corev1.PersistentVolumeClaim, map[string]bool, error) {
    // ... hintOnly[pvcName] = !fromSpec[pvcName] for every PVC added
    return managedPVCs, hintOnly, nil
}

// extractRequestedSize: instead of failing the whole request,
// skip the hint PVC and log why.
if err := resolve(pvc); err != nil {
    if hintOnly[pvc.Name] {
        log.Warning(fmt.Sprintf("skipping PVC %s/%s listed only in the %s annotation: %v",
            pvc.Namespace, pvc.Name, consts.PodExtraPVCsAnnotation, err))
        continue
    }
    return nil, err
}
```

The test that would pin this is the probe above: annotation on a PVC with a broken SC → `filter` returns all incoming nodes.

images/sds-common-scheduler-extender/pkg/consts/consts.go
```diff
@@ -23,6 +23,14 @@ const (
 	LvmTypeParamKey         = "local.csi.storage.deckhouse.io/lvm-type"
 	LVMVolumeGroupsParamKey = "local.csi.storage.deckhouse.io/lvm-volume-groups"
 
+	// PodExtraPVCsAnnotation lists, as a comma-separated value, the names of
+	// PersistentVolumeClaims a Pod needs taken into account when scheduling but
+	// does not mount itself (e.g. a KubeVirt hotplug disk, which is attached by a
+	// separate attachment Pod pinned to the launcher's node). The PVCs are always
+	// looked up in the Pod's own namespace. It is a best-effort scheduling hint:
+	// unknown or foreign-provisioner names are ignored.
+	PodExtraPVCsAnnotation = "scheduler.deckhouse.io/extra-pvcs"
```
fastrapier, line 38: **Severity: High** — The ordering contract is not pinned: if the PVC does not exist yet, the feature silently becomes a no-op

**Tier: Contract must — the producer of this annotation lives in the virtualization module, outside this repository**

The [comment in `getManagedPVCsFromPod`](https://github.com/deckhouse/sds-node-configurator/blob/beebd0b526ae959f6334ca0f7c8ed40a37e1d9e0/images/sds-common-scheduler-extender/pkg/scheduler/func.go#L501-L508) explicitly allows for the PVC to "legitimately not exist yet", and in that case the name is skipped. But that case is not a harmless detail — it is the **primary failure mode of the whole feature**: if the hotplug PVC has not been created by the time the launcher Pod is scheduled, the extender decides blindly again and Kaiten US 67033244 stays unfulfilled.

**Failure mode:** there is no observable difference between "the feature worked" and "the feature stayed silent" — the Pod schedules either way, and the only difference is one `Warning` line in the extender's log.

Who guarantees the "PVC created → launcher Pod created" order is not stated in the PR, and the producer of the annotation lives in another repository and can change independently. Meanwhile the declared scope (`hotplug-at-VM-start`) is the only moment `filter` is called for this launcher at all: there is no second chance.

The key is marked in the description as not yet agreed with the other consumers — so now is exactly the moment to pin the second half of the contract, not just the key string.

**Recommendation:** state the producer requirement in this doc comment (the only place in this repository that consumers read) and put it into the key agreement:

```go
// PodExtraPVCsAnnotation ...
//
// Contract for the producer: every PVC named here MUST already exist when the
// Pod is created. The extender resolves the names once, during filter/prioritize
// for this Pod; a name that does not resolve is skipped, and there is no second
// attempt -- the hint is silently lost and scheduling falls back to being blind.
PodExtraPVCsAnnotation = "scheduler.deckhouse.io/extra-pvcs"
```

If the ordering cannot be guaranteed on the virtualization side, that changes the design: an explicit upward signal would be needed (for example, rejecting nodes with a clear `FailedNodes` reason instead of skipping quietly) — but that is worth settling before the key is agreed, not after.

images/sds-common-scheduler-extender/pkg/scheduler/extra_pvcs_test.go
```diff
@@ -0,0 +1,393 @@
...
+
+	req := httptest.NewRequest(http.MethodPost, "/scheduler/filter", bytes.NewReader(body))
+	w := httptest.NewRecorder()
+	s.filter(w, req)
+
+	require.Equal(t, http.StatusOK, w.Code, "filter must answer 200; body: %s", w.Body.String())
+	var result ExtenderFilterResult
+	require.NoError(t, json.Unmarshal(w.Body.Bytes(), &result))
+	return result
+}
```
fastrapier, line 267: **Severity: Medium** — The `prioritize` handler has no annotation test at all

**Tier: Strong recommendation — Google Testing on the Toilet / the coverage pattern already established in this package (`prioritize_test.go`, `filter_test.go`)**

The description says: "Both handlers (`filter`, `prioritize`) inherit it for free". The first half of that claim is checked by three tests driving the real `/filter`; the second half is not checked at all.

And it is not entirely "free": `prioritize` is a separate handler with its own error branch (`http.Error(500)` instead of `writeFailAllNodesResponse`, see the `func.go` comment) and its own cache step — `narrowReservationsToFinalNodes`. For an annotation-only PVC that step behaves differently than for a spec PVC (see finding 4 in the review body). Precisely the places a test would catch are the ones left uncovered.

**Failure mode:** a future change to `prioritize`'s PVC handling or to the narrowing path silently stops honouring the annotation, and CI stays green because only `filter` is exercised.

The infrastructure is already here: `callFilter` only needs a twin pointing at `/scheduler/prioritize`, and the cluster objects can be reused from `TestFilter_AnnotationPVC_LocalRejectsNodeWithoutSpace`.

**Recommendation:** add `TestPrioritize_AnnotationPVC_ScoresNodes` — a `virt-launcher`-shaped Pod (empty `spec.volumes`, annotation on a pending local PVC), two nodes with different free space, asserting that (a) the response contains both nodes with non-zero scores, (b) the node with more free space scores no lower, (c) the reservation exists after `prioritize` and is narrowed to the pools of the final node list.

```go
func callPrioritize(t *testing.T, s *scheduler, args ExtenderArgs) HostPriorityList {
    t.Helper()
    body, err := json.Marshal(args)
    require.NoError(t, err)
    req := httptest.NewRequest(http.MethodPost, "/scheduler/prioritize", bytes.NewReader(body))
    w := httptest.NewRecorder()
    s.prioritize(w, req)
    require.Equal(t, http.StatusOK, w.Code, "prioritize must answer 200; body: %s", w.Body.String())
    var result HostPriorityList
    require.NoError(t, json.Unmarshal(w.Body.Bytes(), &result))
    return result
}
```

images/sds-common-scheduler-extender/pkg/scheduler/func.go
```diff
@@ -422,50 +423,128 @@ func getNodeNames(inputData ExtenderArgs) ([]string, error) {
...
+			}
 		}
+
+		pvcLog.Debug("[getManagedPVCsFromPod] add PVC to the managed PVCs")
+		managedPVCs[pvcName] = pvc
+	}
+
+	if len(missingAnnotationPVCs) > 0 {
+		log.Warning(fmt.Sprintf("[getManagedPVCsFromPod] PVC name(s) listed in the %s annotation do not exist and were skipped (the annotation is a scheduling hint, not a guarantee): %v", consts.PodExtraPVCsAnnotation, missingAnnotationPVCs))
 	}
```
fastrapier: **Severity: Low** — The aggregated `Warning` is lost if the loop exits early with an error

**Tier: Trade-off**

`missingAnnotationPVCs` is printed after the loop, and the loop has three `return nil, err` exits. If the annotation references a non-existent PVC and the next PVC from `spec.volumes` then errors, the annotation diagnostics never reach the log — even though the operator will be debugging both problems at once. The per-PVC lines survive, but they are at `Debug`, i.e. unavailable at the default log level.

**Failure mode:** a stale annotation plus any unrelated PVC error, and the only clue about the stale annotation disappears from the logs at the moment it is most needed.

**Recommendation:** move the emission into a `defer` right after the slice is declared, so it no longer depends on the exit path:

```go
var missingAnnotationPVCs []string
defer func() {
    if len(missingAnnotationPVCs) > 0 {
        log.Warning(fmt.Sprintf("[getManagedPVCsFromPod] PVC name(s) listed in the %s annotation "+
            "do not exist and were skipped (the annotation is a scheduling hint, not a guarantee): %v",
            consts.PodExtraPVCsAnnotation, missingAnnotationPVCs))
    }
}()
```

images/sds-common-scheduler-extender/pkg/scheduler/func.go
```diff
@@ -422,50 +423,128 @@ func getNodeNames(inputData ExtenderArgs) ([]string, error) {
...
+				continue
+			}
 		}
+
+		pvcLog.Debug("[getManagedPVCsFromPod] add PVC to the managed PVCs")
+		managedPVCs[pvcName] = pvc
+	}
+
+	if len(missingAnnotationPVCs) > 0 {
+		log.Warning(fmt.Sprintf("[getManagedPVCsFromPod] PVC name(s) listed in the %s annotation do not exist and were skipped (the annotation is a scheduling hint, not a guarantee): %v", consts.PodExtraPVCsAnnotation, missingAnnotationPVCs))
```
fastrapier: **Severity: Low** — The same `Warning` is printed on every `filter`/`prioritize` call while the Pod sits `Pending`

**Tier: Trade-off**

Commit `8130129` removed the "one line per PVC" flood — the right move — but the "one line per call" dimension remains. A Pod with a typo in the annotation that cannot be scheduled for some other reason is retried by kube-scheduler with backoff indefinitely, and every attempt costs one line from `filter` and one from `prioritize`. `Warning` is visible at the default log level, so this is noticeable noise, and it grows linearly with time rather than with the number of Pods.

**Failure mode:** a single permanently mis-annotated Pod produces unbounded log growth in the extender for as long as it stays `Pending`.

This is admittedly a trade-off: dropping to `Debug` removes the noise but also removes the typo diagnostics from the default output — and that is the valuable part here.

**Recommendation:** my preference is to keep `Warning` but deduplicate on (Pod UID, sorted list of skipped names) in a small LRU next to `Cache`, so the same complaint is printed once per Pod. If such a cache feels more expensive than the problem, leaving it as is is fine — but then it is worth a comment saying the per-call repetition is deliberate, so it does not read as an unfinished fix of the previous commit.

images/sds-common-scheduler-extender/pkg/scheduler/extra_pvcs_test.go
```diff
@@ -0,0 +1,393 @@
...
+			StorageClassName: &scName,
+			Resources: corev1.VolumeResourceRequirements{
+				Requests: corev1.ResourceList{
+					corev1.ResourceStorage: *resource.NewQuantity(size, resource.BinarySI),
+				},
+			},
+		},
+		Status: corev1.PersistentVolumeClaimStatus{Phase: corev1.ClaimPending},
+	}
+}
```
fastrapier: **Severity: Nit** — `pendingPVCWithSize` duplicates `testPendingPVC` and does not live in the helpers file

`pendingPVCWithSize` is `testPendingPVC` with a parameterised size; the bodies are identical. Commit `8130129` already cleaned up two duplicated constants in this same file — this one stayed. On top of that, a shared helper sits in a specific test's file rather than in [`handler_test_helpers_test.go:390`](https://github.com/deckhouse/sds-node-configurator/blob/beebd0b526ae959f6334ca0f7c8ed40a37e1d9e0/images/sds-common-scheduler-extender/pkg/scheduler/handler_test_helpers_test.go#L390) where all the others live, and the constant is named `extraPVCsTestTenGiB` next to neighbours `hundredGiB` / `oneGiB` from `space_test.go`.

**Recommendation:** move `pendingPVCWithSize` into `handler_test_helpers_test.go`, express the existing helper through it, and rename the constant to match its neighbours:

```go
// handler_test_helpers_test.go
func testPendingPVC(name, namespace, scName string) *corev1.PersistentVolumeClaim {
    return pendingPVCWithSize(name, namespace, scName, oneGiB)
}
```

and `tenGiB` instead of `extraPVCsTestTenGiB`, next to `hundredGiB` / `oneGiB`.

duckhawk (approved):

PR #352, opened 2026-08-14
test(e2e): control run of the dvp suite at merge-base 878512f (do not merge)

duckhawk:
Throwaway control run. **Do not merge; close after the run.** The tree is byte-identical to `main` — the single commit is empty and exists only because the e2e workflow is gated on `pull_request` labels.

Purpose: #351's dvp e2e run fails with `no new consumable block device on node master-1 yet` — 12 occurrences in one attempt and 54 in another, on the same commit. That message appears in neither of the 2026-08-09 dvp runs on `v0-7-0-changelog`, although 7 of the 8 failing spec files already existed there. This run establishes whether the symptom reproduces without #351, separating a regression in the PR from a degraded stand.

Read-out:
- 0 occurrences of `no new consumable block device` here → the symptom belongs to #351, and its unnamed-PV requeue path is the thing to fix.
- The symptom reproduces here → the stand is at fault and #351 is clear.

Context: the `Force removing N stuck LVMVolumeGroups` pattern is *not* a discriminator — it appears at the same magnitude in the 2026-08-09 `main` runs (43/14, 62/21) as in the #351 runs (43/14, 40/14).


duckhawk:
Control run finished: the symptom reproduces on `main` itself.

`main` @ 878512f, zero PR changes: **18 passed / 20 failed**, 48 occurrences of `no new consumable block device on node master-1`.
#351 @ 72f8ab8: 15/23 with 54 occurrences, and 36/5 with 12 occurrences on an earlier attempt of the same commit.

Same failing specs, same single symptom, same node. #351 is not the cause — the dvp stand is. Closing; opened only as a control.

